PR #4, opened 2018-08-23
Support for lang-LOCALE

lchudinov:
* support for lang-LOCALE
 * ability to fall back to lang if lang-LOCALE is not found

Signed-off-by: Leonty Chudinov <lchudinov@rocketsoftware.com>


rpenny125:
I agree. Please fix the code in 
zlux-app-manager\virtual-desktop\src\app\plugin-manager\plugin-manager.module.ts
and
zlux-app-manager\virtual-desktop\src\app\shared\language-locale.service.ts. 

I didn't write such nice fallback code as what you wrote for the translation providers, but I think these are the right places to do the fallback that you describe. After that, the rest of my changes should be OK, right?

You might be able to unify that logic into the language-locale.services.ts since plugin-manager.module.ts already imports language-locale.service.ts

lchudinov:
@rpenny125 I added a fallback for a file with locale data. The rest of your changes are good. Please check my changes with your browser where you have en-US.

PR #5, opened 2018-08-23
Fix ability to set Inputs and subscribe to Outputs of embedded plugins

lchudinov:


1000TurquoisePogs:
Previously reviewed, seems fine.

PR #28, opened 2018-09-27
Reorganized close listener for Apps to be a viewport event

1000TurquoisePogs:
rather than a window event, as a window may have multiple viewports


1000TurquoisePogs:
Leonty said he checked this (in an email) and it seemed good, and checks passed, so I will merge.

PR #31, opened 2018-10-03
Internationalization of desktop controls

lchudinov:


virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-icon/launchbar-icon.component.ts
```diff
@@ -23,8 +23,9 @@ export class LaunchbarIconComponent {
   @Input() launchbarItem: LaunchbarItem;
 
   @Output() iconClicked: EventEmitter<void>;
-
-  constructor(@Inject(MVDHosting.Tokens.ApplicationManagerToken) private applicationManager: MVDHosting.ApplicationManagerInterface) {
+  private applicationManager: MVDHosting.ApplicationManagerInterface;
```
rpenny125: I'm curious why this change is necessary

rpenny125:
This looks promising. Mostly it seems like a fairly limited refactoring, which is comforting. I'm curious about the switch from @Inject to this.injector.

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-icon/launchbar-icon.component.ts
lchudinov: The ngc complained "Can't resolve all parameters for LaunchbarIconComponent..." because ApplicationManagerInterface is under a namespace.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -52,14 +52,18 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
 
   contextMenuRequested: Subject<{xPos: number, yPos: number, items: ContextMenuItem[]}>;
   readonly windowDeregisterEmitter: Subject<MVDWindowManagement.WindowId>;
+  private applicationManager: MVDHosting.ApplicationManagerInterface;
+  private viewportManager: MVDHosting.ViewportManagerInterface;
+  private pluginManager: MVDHosting.PluginManagerInterface;
 
   constructor(
-    @Inject(MVDHosting.Tokens.ApplicationManagerToken) private applicationManager: MVDHosting.ApplicationManagerInterface,
```
1000TurquoisePogs: I looked at this with Robert and now we understand the problem a little more. At first we didn't understand why sample-angular-app could use Inject in i18n, but then realized it was because the Injection items were defined within the angular modules that are importable from virtual-desktop. But that here, MVDHosting is a declared namespace and not a module, so the reference to it is somehow not acceptable to the i18n compiler.

Leonty, although we were able to confirm this behavior, the solution you came up with still bothers me because it sounds like it is avoiding fixing a bigger problem that we have. Can you do another day of research to determine if this problem is caused by a bad structure in zlux-platform, our tsconfig files, or can be fixed by using a different internationalization approach (Robert stated there were 2 angular internationalization libraries we were working with?)

It may be fine to merge this code, but first I want to know what we need to do to avoid more of this in the future, or if we can fix it easily now, because the rest of the codebase is using Inject and it seems more standardized to do that.

virtual-desktop/src/assets/i18n/messages.ja.json
```diff
@@ -0,0 +1,8 @@
+{
+  "Unpin from taskbar": "アプリケーションのピン留めをはずす",
+  "Pin to taskbar": "アプリケーションをピン留めする",
+  "Bring to Front": "最前面に移動",
```
1000TurquoisePogs: Where did you get the Japanese translations from? Just making sure that these are correct.

jplinardon: I got the translations from the dev team in Japan.

virtual-desktop/src/app/i18n/language-locale.service.ts
```diff
@@ -54,7 +52,12 @@ export class LanguageLocaleService {
 
   checkForLocaleFile(localeId: string): Observable<any> {
     const uri = `${this.makeLocaleURI(localeId)}.js`;
-    return this.http.get(uri);
+    return fromPromise(window.fetch(uri).then(res => {
```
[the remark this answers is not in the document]
lchudinov: Thanks for noticing. I fixed this. An import for mergeMap was missing. Note that I use LanguageLocaleService before bootstrap time when Angular Http is not constructed yet.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
lchudinov: I tried various workarounds but they didn't work. If we want to use `@Inject()` then we need to remove the namespace and also add an import statement. AoT compiler doesn't get along with namespaces: https://github.com/angular/angular/issues/15613
Only native Angular i18n approach(with AoT-compiler) is suitable for zlux-widgets and zlux-grid so the plugins that uses zlix-widgets and zlux-grid have to be AoT-compatible. Moreover, in next version of Angular,  i18n-support is going to be significantly improved so AoT-compatibility is important for future.

rpenny125: I take it the workaround provided in that issue did not help? I'm referring to https://github.com/angular/angular/issues/15613#issuecomment-391485560

virtual-desktop/src/app/i18n/language-locale.service.ts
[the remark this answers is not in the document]
rpenny125: This is the kind of thing that warrants a comment. Please add a comment that explains 
1. This code is called before Angualr Http is constructured
2. How it is that this is true.

virtual-desktop/src/app/i18n/translation-loader.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (this.languageLocaleService.isConfiguredForDefaultLanguage()) {
+      return Promise.resolve(noProviders);
+    }
+    const baseLanguage = this.languageLocaleService.getBaseLanguage();
+    // ex.: messages.es-ES.xlf
+    const translationFileURL = this.getTranslationFileURL(plugin, language);
+    // ex.: messages.es.xlf
+    const fallbackTranslationFileURL = baseLanguage !== language ? this.getTranslationFileURL(plugin, baseLanguage) : null;
+    return this.loadTranslations(translationFileURL)
+      .catch(err => (fallbackTranslationFileURL != null) ? this.loadTranslations(fallbackTranslationFileURL) : Observable.throw(err))
```
rpenny125: Please issue a WARNING log message when "falling back":
`Failed to load language file ${ translationFileURL }, using ${fallbackTranslationFileURL }. Error ${err.something()}`

lchudinov: I added such warnings. Please take a look at https://github.com/lchudinov/zowe-zlux-app-manager/commit/6251903aef6e329066fc240206125ad258777474.

lchudinov: I added such warnings. Please take a look at https://github.com/lchudinov/zowe-zlux-app-manager/commit/6251903aef6e329066fc240206125ad258777474.

lchudinov: I added such warnings. Please take a look at https://github.com/lchudinov/zowe-zlux-app-manager/commit/6251903aef6e329066fc240206125ad258777474.

lchudinov: I added such warnings. Please take a look at https://github.com/lchudinov/zowe-zlux-app-manager/commit/6251903aef6e329066fc240206125ad258777474.

lchudinov: I added such warnings. Please take a look at https://github.com/lchudinov/zowe-zlux-app-manager/commit/6251903aef6e329066fc240206125ad258777474.

lchudinov: I added such warnings. Please take a look at https://github.com/lchudinov/zowe-zlux-app-manager/commit/6251903aef6e329066fc240206125ad258777474.

1000TurquoisePogs:
Did a smoke test where I put my browser in Japanese and it looks like it fits the requirements, so I'm merging.

PR #39, opened 2018-10-26
add a property window that can popup

BaoyuanXing:
refactor the app manager and window manager, which would allow future development to launch multiple instances of plugin.
add copyright field in pluginDefinition for display purpose
rename plguin id of virtualdeskop and bootstrapper in zowe naming conventioin.

Signed-off-by: Baoyuan Xing <bxing@rocketsoftware.com>


BaoyuanXing:
this is a cleaner branch. The old one would be deleted after merging.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -239,16 +233,18 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
...
     return windowId;
   }
-
+   
+ 
   getWindow(plugin: MVDHosting.DesktopPluginDefinition): MVDWindowManagement.WindowId | null {
-    const desktopWindows = this.pluginMap.get(plugin.getIdentifier());
-    if (desktopWindows !== undefined) {
-      return desktopWindows[0].windowId;
+    const applicationWindowID = this.applicationManager.getAppWindowID(plugin.getIdentifier());
```
1000TurquoisePogs: Why did you move windowID decision making outside of windowmanager service and into applicationmanager? window manager manages windows, application manager manages applications. applications are spawned into viewports, which the viewport manager handles. viewports go into windows, which the window manager handles.
Some of the existing code is wrong, but I think this would make it more wrong. I'd rather move all window logic from application manager into window manager. This will make more sense when Suneeth finishes multi-window support per application.

virtual-desktop/src/app/application-manager/application-manager.service.ts
```diff
@@ -217,17 +237,60 @@ export class ApplicationManager implements MVDHosting.ApplicationManagerInterfac
...
+    };    
+  }
+  
+  showApplicationPropertiesWindow(plugin: DesktopPluginDefinitionImpl): void {
     const windowManager: MVDWindowManagement.WindowManagerServiceInterface = this.injector.get(MVDWindowManagement.Tokens.WindowManagerToken);
-    return windowManager.getWindow(plugin) != null;
+    const pluginID = plugin.getIdentifier();
+    let windowId = this.pluginPropertyShowingMap.get(pluginID);
+    if (windowId==undefined){
+      this.pluginManager.findPluginDefinition("org.zowe.zlux.appmanager.app.propview").then(viewerPlugin => {
```
1000TurquoisePogs: The application manager shouldn't be hardcoding this. The application manager should only spawn an application when given an application ID, via spawnapplication. See my longer comment in https://github.com/zowe/zlux-platform/pull/7

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
BaoyuanXing: I thought either app manager or window manager should be more powerful on something to make it possible to launch multiple instances of one plugin. So I moved some of window manager part into app manager, which made window manager simpler and app manager more complicated. Sure I could do verse vice.

BaoyuanXing:
Hi Sean,
You probably checked out the old branch. This branch name is feature/property-window. The old one should be removed.

1000TurquoisePogs:
Yes, looks like I was on the wrong branch somehow.

PR #42, opened 2018-11-08
MVD-2139 Add the version to the URL

fkovinAtRocket:
MVD-2139 The URI broker in the client needs to be updated
Signed-off-by: Fyodor Kovin fkovin@rocketsoftware.com


bootstrap/src/uri/dsm-uri.ts
```diff
@@ -113,23 +113,27 @@ export class DsmUri implements ZLUX.UriBroker {
    */
   pluginConfigForScopeUri(pluginDefinition: ZLUX.Plugin, scope: string, resourcePath:string, resourceName?:string): string {
```
1000TurquoisePogs: Add version as an actual optional parameter, but default to _current. Give people that option to specify. In the future, the uriBroker can be clever in taking different parameters or handling them differently according to versions.

fkovinAtRocket:
Anyway, here's a new pull request https://github.com/zowe/zlux-platform/pull/10

bootstrap/src/uri/dsm-uri.ts
1000TurquoisePogs: I'd think including the config service, actually, but we can probably get away with _current for it for now too.
Yes, everything that could have _current can also have a specific version, and therefore it should be a parameter, even if its unlikely people would use it.

PR #46, opened 2018-11-12
Renaming plugins to be org.zowe

1000TurquoisePogs:


system-apps/app-prop-viewer/pluginDefinition.json
```diff
@@ -15,7 +15,9 @@
     "descriptionDefault": "Zowe Viewer",
     "defaultWindowStyle": {
       "width": 350,
-      "height": 450
+      "height": 450,
+      "x": 0,
```
rpenny125: I thought we didn't support x and y here anymore. I know other plugins have this, but it's ignored. Did it somehow get into the json schema? Why are we adding it to more places when we aren't using it?

1000TurquoisePogs: It's in the ticket for using json-schema, so I was adding it.

rpenny125: We can leave it in here for now. However, we should remove the x, y coordinates from the json schema, as the fields are not currently used.

At best, x, y could be made optional.

PR #47, opened 2018-11-15
Add sonar properties

MarkAckert:
These changes are minimal viable configurations to allow sonar to scan zlux and its submodules.

See https://github.com/zowe/zlux/pull/50

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,4 @@
+sonar.projectKey=zlux:zlux-app-manager
+sonar.projectName=ZLUX Application Manager
+sonar.projectVersion=1.0.0
+sonar.sources=virtual-desktop/nodeServer,virtual-desktop/src,bootstrap/src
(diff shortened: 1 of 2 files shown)


sonar-project.properties
```diff
@@ -0,0 +1,4 @@
+sonar.projectKey=zlux:zlux-app-manager
+sonar.projectName=ZLUX Application Manager
+sonar.projectVersion=1.0.0
+sonar.sources=virtual-desktop/nodeServer,virtual-desktop/src,bootstrap/src
```
1000TurquoisePogs: Also, /system-apps/*/webClient, /system-apps/*/nodeServer (potentially) are needed, due to https://github.com/zowe/zlux-app-manager/tree/master/system-apps/app-prop-viewer

sonar-project.properties
```diff
@@ -0,0 +1,4 @@
+sonar.projectKey=zlux:zlux-app-manager
+sonar.projectName=ZLUX Application Manager
+sonar.projectVersion=1.0.0
```
1000TurquoisePogs: What is the meaning of the version? zlux is 0.9.3 but this is 1.0.0. In reality this is a core repo to zlux so I'd expect them to be the same.

sonar-project.properties
MarkAckert: I added properties for the system-apps I could pull down from master and changed the project version to 0.9.4 for now - longer term pulling the version from the source would be preferred. This is intended to be MVP - something that runs and you can check the output for.

PR #49, opened 2018-11-16
My local product specific changes

JerryGaoRocket:
These changes are for my local product, I don't think these should merge into the master branch of base repo.


1000TurquoisePogs:
Thanks for opening this, but it looks like the changes here are not something we want to do in Zowe at this time. Rather, we should instead do the following:
1. Increase password length limit to 100 (this is what the mainframe uses, and ultimately server-side checks on length should still be done to reject if, in your case only 48 is the maximum). I made another PR for this.
2. Find another means to avoid error messages when App Components aren't found, as this is an optional, not a required, ability of apps.

PR #51, opened 2018-11-27
Use requirejs 0 second timeout to avoid high latency timeouts

1000TurquoisePogs:
If you are on a high latency connection, the Desktop or Apps can fail to load due to requirejs having a default timeout value of 7 seconds. That is, if a file doesnt load in 7 seconds, requirejs gives up on it and leaves you with a dead page. See https://github.com/requirejs/requirejs/issues/787 and https://requirejs.org/docs/api.html#config-waitSeconds

Setting the timeout value to 0 disables it. This commit does not include any new way to indicate to the users that the page is still loading, but will allow the page to load where it would previously not.

### DIFF
--- a/virtual-desktop/src/externals.ts
+++ b/virtual-desktop/src/externals.ts
@@ -13,7 +13,6 @@
 /*
   Imperfect solution to an imperfect world. See https://github.com/requirejs/requirejs/issues/787 and https://requirejs.org/docs/api.html#config-waitSeconds
 */
-console.log("I set that config");
 (window as any).requirejs.config({
   waitSeconds: 0
 });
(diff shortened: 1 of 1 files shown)


virtual-desktop/src/externals.ts
```diff
@@ -10,6 +10,14 @@
   Copyright Contributors to the Zowe Project.
 */
 
+/*
+  Imperfect solution to an imperfect world. See https://github.com/requirejs/requirejs/issues/787 and https://requirejs.org/docs/api.html#config-waitSeconds
+*/
+console.log("I set that config");
```
lchudinov: Please remove a call to console.log.

1000TurquoisePogs: Whoops, thanks!

1000TurquoisePogs:
@DivergentEuropeans when the timeout is reached, the page actually breaks whereas it may not have if you were patient. So, I think any number is bad, but a loading screen would be nice.

PR #52, opened 2018-11-28
Cascading windows/Remembering window positions

DivergentEuropeans:
1. This PR solves the bug where opening and closing the same plugin will move the newly opened window farther and farther down the page (in a cascade-type fashion). This was caused by a cascade logic that relied on "last opened window", which can be unreliable at times (for ex. if the last opened window happens to be the same plugin).

Instead, the cascade now works by using an index, so 
for ex. if I opened plugins A, B, C, 

      ___ A ________
     |        ___ B ________
     |       |        ___ C ________ 
             |       |
                     |

then closed B

      ___ A ________
     |        
     |                ___ C ________ 
                     |
                     |

then opened plugin D, I would get

      ___ A ________
     |        ___ D ________
     |       |        ___ C ________ 
             |       |
                     |
As each plugin will move down via cascade, with plugins filling up available slots.

2. This PR makes it so that Zowe remembers the positions of user's previously opened plugins. So if a user last closed a plugin at a certain position, with a certain size, it will pop back up with that same size in the same desired position.

In case, the plugin window is too large to move farther down the cascade, the behavior is the same as prior to the PR.

### DIFF
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
@@ -106,13 +106,16 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
     let { width: dtWindowWidth, height: dtWindowHeight } = plugin.defaultWindowStyle;
     const rightMostPosition = innerWidth - dtWindowWidth;
     const bottomMostPosition = innerHeight - dtWindowHeight;
-
+    let desktopHeight = document.getElementsByClassName('window-pane').item(0).getBoundingClientRect().bottom;
+    let desktopWidth = document.getElementsByClassName('window-pane').item(0).getBoundingClientRect().right;
+    let launchbarHeight = document.getElementsByClassName("launchbar-container")[0].clientHeight;
     const pluginId = plugin.getIdentifier();
     if (this.windowPositionsMap.has(pluginId))//If a plugin has previously saved window position & size data
     { return {
         ...this.windowPositionsMap.get(pluginId),
       } as WindowPosition
     }
+    /*
 
     let valueArray = Array.from(this.runningPluginMap.values());
     let missingIndex = 0;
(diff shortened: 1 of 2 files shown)


virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -50,7 +50,7 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
    * 2. Even though this looks like it starts at (0.0), doing this simplifies the logic of
    *    generateNewWindowPosition whuile still starting off at (NEW_WINDOW_POSITION_INCREMENT, NEW_WINDOW_POSITION_INCREMENT)
    */
-  private lastWindowPosition: WindowPosition;
+  private windowPositionsMap: Map<String, WindowPosition>;
```
1000TurquoisePogs: This will be a problem as soon as anyone tries to use 2 instances of the same app.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -106,9 +107,26 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
...
-    let nextLeft = lastTopLeft + WindowManagerService.NEW_WINDOW_POSITION_INCREMENT;
-    let nextTop = lastTopRight + WindowManagerService.NEW_WINDOW_POSITION_INCREMENT;
+    const pluginId = plugin.getIdentifier();
+    if (this.windowPositionsMap.has(pluginId))//If a plugin has previously saved window position & size data
+    { return {
+        ...this.windowPositionsMap.get(pluginId),
+      } as WindowPosition
+    }
+
+    let valueArray = Array.from(this.runningPluginMap.values());
```
1000TurquoisePogs: Why are you cascading indiscriminately? If you are trying to mimic other desktops, they do not do this. Rather, they group instances of each program. So, on Windows for example, it will remember that the last instance of Chrome was at offset 2,55 and will cascade new instances of Chrome from there. However, it will have no bearing on notepad, which may begin cascading from 77,6 if that were its last position.

DivergentEuropeans: Because that's how Zowe had it now with its indescriminant last window position cascade. It didn't matter what category of applications there were. But I can probably make it so that it does though. Will just require a bit of work, I'll get back to you.

PR #56, opened 2018-11-29
Fix for ExpressionChangedAfterItHasBeenCheckedError in login dialog.

lchudinov:
Add additional change detection. See for reference: https://github.com/angular/angular/issues/22426

Steps to reproduce the error:
* Sign out and re-load the page with `Ctrl+F5`
* Enter username and password, and **press `Enter`**


1000TurquoisePogs:
Apparently this is chrome-specific and the fix could likewise be conditional but since it doesn't hurt, it's fine unconditionally too just in case other browsers have a change in behavior.

PR #60, opened 2018-12-03
components.js not requested if plugin definition does not specify hasComponents

reetchow:
The components.js file is no longer requested from the server unless the plugin specifies that it uses Zowe components. This has the benefit of not sending network requests that result in a 404.

### DIFF
--- a/virtual-desktop/src/app/plugin-manager/plugin-factory/angular2/angular2-plugin-factory.ts
+++ b/virtual-desktop/src/app/plugin-manager/plugin-factory/angular2/angular2-plugin-factory.ts
@@ -107,24 +107,30 @@ export class Angular2PluginFactory extends PluginFactory {
     const scriptUrl = Angular2PluginFactory.getAngularComponentsURL(pluginDefinition);
 
     return new Promise((resolve, reject) => {
-      (window as any).require([scriptUrl],
-        (components: MvdNativeAngularPluginComponentDefinition) => {
-          const factoryDefs = components.getComponentFactoryDefinitions(pluginDefinition);
-          factoryDefs.forEach((factory: AngularComponentFactoryDefinition) => {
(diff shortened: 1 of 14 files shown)


virtual-desktop/src/app/plugin-manager/plugin-factory/angular2/angular2-plugin-factory.ts
```diff
@@ -107,24 +107,30 @@ export class Angular2PluginFactory extends PluginFactory {
...
+          },
+          (failure: any) => {
+            console.log(`No component definition for plugin ${pluginDefinition.getIdentifier()}`);
             resolve();
           });
-        },
-        (failure: any) => {
-          console.log(`No component definition for plugin ${pluginDefinition.getIdentifier()}`);
+        } else {
+          console.log('This plugin does not use Zowe components');
```
1000TurquoisePogs: This needs to be more descriptive, otherwise you will get tons of "this plugin does not ..." and you'll be thinking "which one"?
Try:
`Plugin (${pluginDefinition.getIdentifier()}) didn't expose Zowe components. Component loading skipped`

virtual-desktop/src/app/plugin-manager/shared/desktop-plugin-definition.ts
```diff
@@ -38,7 +38,12 @@ export class DesktopPluginDefinitionImpl implements MVDHosting.DesktopPluginDefi
...
-      return this.basePlugin.getWebContent().framework;
+      if ('framework' in this.basePlugin.getWebContent()) {
+        return this.basePlugin.getWebContent().framework;
+      } else {
+        console.warn(`Plugin ${this.getIdentifier()} has no framework specified`);
+        return 'unsupported';
+      }
     } else {
       console.warn(`Plugin ${this.getIdentifier()} has no framework specified`);
       return 'unsupported';
```
1000TurquoisePogs: If this were to return N/A, then that would make sense as compared to 'unsupported'. Then, no console warn would be necessary.

1000TurquoisePogs:
Redundant

PR #63, opened 2018-12-04
Replace "Authentication failed..." message.

lchudinov:
Replace "Authentication failed for 1 types. Types: ["zss"]" with two messages
* Incorrect user name and/or password
* Failed to communicate with zss authentication server

Test steps:
* stop zss-server and try to login - expected error message: Failed to communicate with zss authentication server
* start zss-server and try to login with wrong credentials - expected error message: Incorrect user name and/or password

This PR depends on https://github.com/zowe/zss-auth/pull/5

### DIFF
--- a/virtual-desktop/src/app/authentication-manager/login/login.component.ts
+++ b/virtual-desktop/src/app/authentication-manager/login/login.component.ts
@@ -63,11 +63,7 @@ export class LoginComponent implements OnInit {
           try {
             const jsonMessage = error.json();
             if (this.isAuthResponseMessage(jsonMessage)) {
-              if (this.isZssConnectionErrorMessage(jsonMessage)) {
-                this.errorMessage = this.translation.translate('FailedToCommunicateWithZssAuthenticationServer');
-              } else {
-                this.errorMessage = this.translation.translate('IncorrectUsernameAndOrPassword');
-              }
+              this.handleAuthResponseMessage(jsonMessage);
             } else {
               this.errorMessage = error.text();
             }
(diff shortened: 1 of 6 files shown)


DivergentEuropeans:
""IncorrectUsernameAndOrPassword": "Incorrect user name and/or password","

Would it be the best way to go to assume that the username or password are incorrect when we get a bounce back?

For ex. if I was using Zowe on some Friday, the LPARs get cleaned over the weekend and my zss server goes down, then I go into log in, it would tell me that my credentials are incorrect. But in reality, my credentials are probably fine, it just can't check because my zss server is down. In a perfect world, you would want to have those scenarios as two separate errors, but I don't know if that case occurs often enough to warrant maybe rewording to something like "Authentication failed or zss server cannot be reached."

virtual-desktop/src/app/authentication-manager/login/login.component.ts
```diff
@@ -135,11 +118,65 @@ export class LoginComponent implements OnInit {
     );
   }
 
+  private handleAuthResponseMessage(authMessage: AuthResponseMessage): void {
+    if (this.isZssConnectionErrorMessage(authMessage)) {
```
1000TurquoisePogs: This improves the UX for ZSS but I don't like that it only works for ZSS. I'd like it even more if "ConnectionError" was a meaningful response for ANY auth handler, so that the end user would know that you couldn't login due to a server connectivity issue whether it was ZSS or another server. Seems to me that the first of any auth category that has ConnectionError should print this error, whether it is zss or not. And I'm sure that the i18n translations don't change grammatically when substituting zss for any other category name.

PR #65, opened 2018-12-05
Feature for multiple instances of applications

sakeerthy:

### DIFF
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-instance-view/launchbar-instance-view.component.ts
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-instance-view/launchbar-instance-view.component.ts
@@ -26,8 +26,20 @@ export class LaunchbarInstanceViewComponent {
   }
 
   ngOnInit(){
-    (<HTMLImageElement>document.getElementsByClassName("instance-viewer")[0]).style.left = 
-      -(180 + (105  * (this.launchbarItem.windowPreviewsIds.length - 2))) + 'px';
+    //Doesn't check right bounds yet, for somereason bounds.right isn't giving the expected values
+    //and not sure why at this time
+    this.launchbarItem.showIconLabel = false;
+    let bounds = (<HTMLImageElement>document.getElementsByClassName("instance-viewer")[0]).getBoundingClientRect();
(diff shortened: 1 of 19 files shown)


sakeerthy:
This PR is also dependent on the PR in zlux-platform which can be found at https://github.com/zowe/zlux-platform/pull/15

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-instance-view/launchbar-instance-view.component.ts
```diff
@@ -26,8 +26,20 @@ export class LaunchbarInstanceViewComponent {
...
-    (<HTMLImageElement>document.getElementsByClassName("instance-viewer")[0]).style.left = 
-      -(180 + (105  * (this.launchbarItem.windowPreviewsIds.length - 2))) + 'px';
+    //Doesn't check right bounds yet, for somereason bounds.right isn't giving the expected values
+    //and not sure why at this time
+    this.launchbarItem.showIconLabel = false;
+    let bounds = (<HTMLImageElement>document.getElementsByClassName("instance-viewer")[0]).getBoundingClientRect();
+    //260 offset is height of the instance viewer plus the height of the taskbar
+    (<HTMLImageElement>document.getElementsByClassName("instance-viewer")[0]).style.top = (document.body.clientHeight - 250) + 'px';
+    if (bounds != null) {
+      if (bounds.left - (180 + (105  * (this.launchbarItem.windowPreviewsIds.length - 2)))  < 0 ) {
```
1000TurquoisePogs: We always do constants with CAPITALS_AND_UNDERSCORES (at the top of files) to keep track of specific numbers like this, since it makes it easy to change later as well as to be more self-explanatory as to what is actually happening.

1000TurquoisePogs:
I reviewed with Suneeth and we think that there are some bugs that can't be reasonably solved without reorganizing the "task viewer" to be owned by the launchbar, rather than the app icons on the launchbar (as it is now). So, I'd rather get this code in prior to that reorganization so that people can start to depend on it at a higher level, without worrying about the implementation details, and then we can iterate to fix the bugs that we saw over time. The bugs that we will need to fix include that the task viewer shouldnt go away upon mouse leaving, only mouse clicking, as well as that the viewer should always be below the icons (they can get cut off partially as it is now).

PR #69, opened 2018-12-13
Uri was wrong for dsn metadata lookup'

1000TurquoisePogs:


1000TurquoisePogs:
I am merging this since Jordan is involved with the API, and has checked off on this.

PR #73, opened 2019-01-08
Fix Zowe build

DenisGlebov:
Build was failed due to compiler errors:

ERROR in C:\Dev\web-tep-cap\zlux-app-manager\bootstrap\src\uri\dsm-uri.ts
     [exec] [tsl] ERROR in C:\Dev\web-tep-cap\zlux-app-manager\bootstrap\src\uri\dsm-uri.ts(28,15)
     [exec]       TS6133: 'route' is declared but its value is never read.
     [exec]
....etc

+ optional parameters declarations clearing


1000TurquoisePogs:
Tested that this fixes the broken build.

PR #74, opened 2019-01-08
Fix version for copy-webpack-plugin

1000TurquoisePogs:


1000TurquoisePogs:
In https://github.com/zowe/zlux-app-manager/pull/77

PR #83, opened 2019-01-21
Build fix for missing alias

1000TurquoisePogs:
Externals building fails to find virtual desktop logger, due to missing alias. This fixes it


jvandun76:
This fix indeed addresses the build problem, thanks!

PR #87, opened 2019-01-28
Detect session timeout & renew

1000TurquoisePogs:
Depends on https://github.com/zowe/zlux-server-framework/pull/69
Depends on https://github.com/zowe/zlux-widgets/pull/11
Depends on https://github.com/zowe/zss-auth/pull/11
Depends on https://github.com/zowe/zosmf-auth/pull/9

Currently the Zowe Desktop does not know when sessions will expire, so when they do apps misbehave. 
By having the server present information about expiration to the Desktop, the Desktop can act upon this to either renew the session before expiration, or log the user out upon expiration so that either way users get an experience that is consistent with the backend session knowledge.

When it comes to session renewal, it does not make sense to renew the session if the user is idle. Therefore, near session expiration, if the user is idle, it will utilize zlux-widgets (new prereq now) to pop up a warning, that the user can click to signal that they are not idle. Otherwise, if they are idle, then they will get logged out at expiration time, with a message stating that the reason for the logout was session expiration.

TODOs not covered here:
- No logic to detect session expiration in realtime as a result of reading HTTP responses... this is tricky without service workers and an extensible environment
- No intelligence so far on knowing if ZSS & ZOSMF's session length is not the default

### DIFF
--- a/bootstrap/package-lock.json
+++ b/bootstrap/package-lock.json
@@ -25,13 +25,6 @@
       "integrity": "sha512-7EJYyKTL7tFR8+gDbB6Wwz/arpGa0Mywk1TJbNzKzHtzbwVmY4HR9WqS5VV7dsBUKQmPNr192jHr/VpBluj/hg==",
       "dev": true
     },
-    "abbrev": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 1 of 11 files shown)


fkovinAtRocket:
> No logic to detect session expiration in realtime as a result of reading HTTP responses... this is tricky without service workers and an extensible environment

I'd like to explore this possibility though. Can we do this:
1. Write a generic wrapper for the XHR or fetch() that would handle expiration and invoke a callback to refresh the session
2. Write an adaptation for Angular
3. Write an adaptation for React

It may sound like a lot, but I did that a couple times and in fact that was pretty trivial for someone who knows the environment. I would rather investigate those options.

1000TurquoisePogs:
BTW on your note on framework-specific 401-interceptors... we may do this, but have to review as future work and how much it accomplishes versus if this session timeout detection work will make the need to sense 401s unnecessary.

PR #90, opened 2019-02-10
IFrame focus-on-click + Windows get focus via resize handles

DivergentEuropeans:
• This solves the bug where IFrames can't be selected into focused context by clicking inside them, only by clicking on their window header. This is accomplished by syncing the IFrame IDs with their window array counterparts, instead of using an IFrame index that has no real tie-in to the rest of the desktop environment outside IFrames. Then, we can use the window ID of the IFrame obtained from the event an IFrame fires off when clicked on (which is basically no event but we can instead detect when the desktop loses focus aka a blur, and then check if it was an IFrame).

• This also adds focus to windows via their resize handles. In most OS, for ex. Mac, Windows & most variants of Linux, when you click on a resize handle to resize a window, it puts the window up front into focus, in combination with resizing it. If you go to re-size a non-focused window, Zowe will resize it in the background without giving it focus, which is kind of a weird feature. 

### DIFF
--- a/virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin-factory.ts
+++ b/virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin-factory.ts
@@ -34,7 +34,7 @@ export class IFramePluginFactory extends PluginFactory {
     window.addEventListener("blur", (event) => { //Checks if focus is lost from the desktop
       if (iFrameElement != null && document.activeElement.className == "mvd-iframe") //Checks if an IFrame caused it
       {
-        let stringId = iFrameElement.id.replace(/[^0-9\.]+/g, ""); //Extracts the Window ID from the IFrame ID
+        let stringId = iFrameElement.id.replace(/[^0-9\.]+/g, ""); //Extracts the instance ID from the IFrame ID
         const windowManager: MVDWindowManagement.WindowManagerServiceInterface = this.injector.get(MVDWindowManagement.Tokens.WindowManagerToken);
         windowManager.showWindow(parseInt(stringId, 10)); 
       }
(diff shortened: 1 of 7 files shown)


virtual-desktop/src/app/plugin-manager/shared/plugin-loader.ts
```diff
@@ -47,9 +47,8 @@ export class PluginLoader {
     });
   }
 
-  loadPlugin(pluginDefinition: MVDHosting.DesktopPluginDefinition): Promise<CompiledPlugin> {
+  loadPlugin(pluginDefinition: MVDHosting.DesktopPluginDefinition, windowId: MVDHosting.ViewportId): Promise<CompiledPlugin> {
```
1000TurquoisePogs: windowId != viewportId because a window can contain multiple viewports, although this isn't utilized much in the existing apps.

virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin-factory.ts
```diff
@@ -75,8 +86,8 @@ export class IFramePluginFactory extends PluginFactory {
     return Promise.resolve();
   }
 
-  loadPlugin(pluginDefinition: MVDHosting.DesktopPluginDefinition): Promise<CompiledPlugin> {
-    const componentClass = this.createIFrameComponentClass(pluginDefinition);
+  loadPlugin(pluginDefinition: MVDHosting.DesktopPluginDefinition, pluginId: MVDHosting.ViewportId): Promise<CompiledPlugin> {
```
1000TurquoisePogs: You shouldn't call the viewportId pluginId, since pluginDefinition.identifier is to me, also pluginId and thus misleading.

virtual-desktop/src/app/plugin-manager/shared/plugin-loader.ts
DivergentEuropeans: Thanks for pointing that out.

PR #93, opened 2019-03-19
fixed window handles move window for north and west handles

reetchow:
This change adds some conditional logic to check which window handle has been clicked and dragged. If it's the North or West handles then the additional logic does not add to the top left anchor point if the window is compressed below minWidth and minHeight.

Signed-off-by: Reet Chowdhary <rchowdhary@rocketsoftware.com>


dmitrynikolaev:
Jenkins automation has been trying to run all possible unit tests, and in some directories we have bad package.json files without name, version, etc.

Amended to skip such packages.

1000TurquoisePogs:
Looks good. Thanks

PR #95, opened 2019-03-28
Personalization Panel + Language Switcher

DivergentEuropeans:
This PR began from a Rocket Build project and lays the code groundwork & structure for an in-depth customization system for Zowe users that will make way for desktop settings, accessibility options, visual themes & other things we want to create to make the user experience better. More specifically, this PR is concerned with adding Internationalization via a Language switcher.

**Major Changes:**

• Personalization panel - Adds a personalization & settings panel that allows the user to select the Language switcher. This would be populated by personalization settings in the future as they are made.
 
image

• App tray - Adds an icon to access the personalization panel. 

image

• Language switcher - The first module of the larger personalization panel is a Language switcher. This will allow Zowe users to switch to one of the supported languages & locales: English, Chinese, Japanese, Russian, & French.

image

All visuals have been made or approved by UX Timofey & Craig.


virtual-desktop/src/app/window-manager/mvd-window-manager/personalization-panel/personalization-panel.component.css
```diff
@@ -0,0 +1,94 @@
...
+  background-position: center;
+  cursor: pointer;
+  border: none;
+  padding: 70px;
+  -webkit-filter: drop-shadow( 0px 0px 5px #00000063 );
+  filter: drop-shadow( 0px 0px 5px #00000063 );
+}
+
+.panel-item-keyboard{
+  background-image: url('../../../../../../virtual-desktop/src/assets/images/personalization/keyboard.png');
```
1000TurquoisePogs: Is there no better way to do this? It's an extremely long relative path that seems like we will regret eventually.

virtual-desktop/tsconfig.json
```diff
@@ -16,8 +16,8 @@
     "noImplicitAny": true,
     "noImplicitReturns": true,
     "noImplicitThis": true,
-    "noUnusedLocals": true,
-    "strictNullChecks": true,
+    "noUnusedLocals": false,
```
1000TurquoisePogs: Please revert this... they're good checks to have.

system-apps/system-settings-preferences/webClient/package.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "org.zowe.zlux.appmanager.system.settings-preferences",
+  "name": "org.zowe.zlux.ng2desktop.settings",
   "version": "1.0.0",
   "scripts": {
     "start": "webpack --progress --colors --watch",
     "build": "webpack --progress --colors",
-    "i18n": "ng-xi18n -p tsconfig.i18n.json --i18nFormat=xlf  --outFile=messages.xlf",
+    "i18n": "ng-xi18n -p tsconfig.ngx-i18n.json --i18nFormat=xlf  --outFile=messages.xlf && xliffmerge -p xliffmerge.json",
```
1000TurquoisePogs: build fails because xliffmerge doesn't exist, neither does xliffmerge.json
This seems to be for the desktop itself, not apps.

Other apps don't do this:
https://github.com/zowe/sample-angular-app/blob/master/webClient/package.json#L7

1000TurquoisePogs:
Do I know English if I'm trying to switch between Chinese and Russian? Probably but...
image

1000TurquoisePogs:
Looks good now, will merge after rc cut.

PR #97, opened 2019-03-29
websocket uri change for api-layer

reetchow:
This change is to ensure when websocket URIs are requested from behind the api mediation layer that routes which look like /ui/v1/zlux/ get changed to /ws/v1/zlux/. Currently there is no dynamic/programmatic way of getting the websocket endpoints so this will have to suffice for now.

Signed-off-by: Reet Chowdhary <rchowdhary@rocketsoftware.com>

### DIFF
--- a/bootstrap/src/uri/mvd-uri.ts
+++ b/bootstrap/src/uri/mvd-uri.ts
@@ -110,7 +110,9 @@ export class MvdUri implements ZLUX.UriBroker {
     const wsProtocol = (protocol === 'https:') ? 'wss:' : 'ws:';
     const uri = `${wsProtocol}//${window.location.host}${this.pluginRootUri(plugin)}`
         + `services/${serviceName}/${version}/${relativePath}`;
-    return proxy_mode ? uri.replace('ui', 'ws') : uri;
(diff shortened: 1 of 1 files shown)


bootstrap/src/uri/mvd-uri.ts
```diff
@@ -108,8 +108,9 @@ export class MvdUri implements ZLUX.UriBroker {
     }
     const protocol = window.location.protocol;
     const wsProtocol = (protocol === 'https:') ? 'wss:' : 'ws:';
-    return `${wsProtocol}//${window.location.host}${this.pluginRootUri(plugin)}`
+    const uri = `${wsProtocol}//${window.location.host}${this.pluginRootUri(plugin)}`
         + `services/${serviceName}/${version}/${relativePath}`;
+    return proxy_mode ? uri.replace('ui', 'ws') : uri;
```
1000TurquoisePogs: This needs a comment, and a minor change.
It should be '/ui/' with '/ws/'. This is regex so it may need '\/' for slashes.
AND, it needs a comment above, saying why this is being done.
Essentially at this time we know it to be a hardcoded behavior that we must reflect here, so we are making an assumption that proxy_mode = api layer, and api layer's ws always is /ws, and everything else we care about is /ui.

PR #99, opened 2019-04-09
Added German + missing translation files + bugs

DivergentEuropeans:
• Added German
• Added missing translations/modified translations for some. Also, translation files are now in alphabetized order. It was fairly strange not having this before, and it will just become harder and harder to keep track of them without it.
• Fixed minor bug in language selection UI


dmitrynikolaev:
There's a "fake" package in `web/locales` that should've be ignored by the pipeline. Fixed.

PR #101, opened 2019-04-12
Add gzip support for large bootstrapping files

1000TurquoisePogs:
gzip compression is supported by pretty much everything.
The main.js, desktop.js, and externals.js files are large enough that on slow connections, they really slow down the time it takes to load the desktop.
This makes the build process make .gz variants in the /web directory, so that the server can serve them to clients that request them.
Brotli, an even better compression algorithm supported by many browsers, was tested but backed out because the libraries required do not seem to work on z/os at this time. The difference between brotli and gzip is minor compared to gzip versus no compression at all.


1000TurquoisePogs:
Need an additional testing: brotli webpack library uses node-gyp, so it's possible it won't work on z/os for building. If it doesn't, I will need to pull it out and use just gzip. Has no effect on the server's serving of brotli if present.

PR #106, opened 2019-05-22
Init windows to top-left rather than center of screen

1000TurquoisePogs:

### DIFF
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
@@ -121,17 +121,15 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
 
   private generateNewWindowPosition(plugin: DesktopPluginDefinitionImpl): WindowPosition {
     let { width: dtWindowWidth, height: dtWindowHeight } = plugin.defaultWindowStyle;
-    const desktopHeight = document.getElementsByClassName('window-pane')[0].clientHeight;
-    const desktopWidth = document.getElementsByClassName('window-pane')[0].clientWidth;
     const launchbarHeight = WindowManagerService.LAUNCHBAR_HEIGHT;
     const { innerWidth, innerHeight } = window;
     const rightMostPosition = innerWidth - dtWindowWidth;
     const bottomMostPosition = innerHeight - dtWindowHeight;
     const pluginIdentifier = plugin.getIdentifier();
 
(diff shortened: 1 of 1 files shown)


virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -121,17 +121,15 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
 
   private generateNewWindowPosition(plugin: DesktopPluginDefinitionImpl): WindowPosition {
     let { width: dtWindowWidth, height: dtWindowHeight } = plugin.defaultWindowStyle;
-    const desktopHeight = document.getElementsByClassName('window-pane')[0].clientHeight;
```
DivergentEuropeans: Can we keep this commented out instead of deleted? Since maybe we'll come back to it (myself, you, or maybe an intern for example) and get a good center-of-screen app spawning method going.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -121,17 +121,15 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
...
     const launchbarHeight = WindowManagerService.LAUNCHBAR_HEIGHT;
     const { innerWidth, innerHeight } = window;
     const rightMostPosition = innerWidth - dtWindowWidth;
     const bottomMostPosition = innerHeight - dtWindowHeight;
     const pluginIdentifier = plugin.getIdentifier();
 
-    // By default, plugins begin in the center of the screen (half of both x & y axes)
-    let nextLeft = (desktopWidth / 2) - (dtWindowWidth / 2);
-    let nextTop = (desktopHeight / 2) - (dtWindowHeight / 2) - (WindowManagerService.WINDOW_HEADER_HEIGHT / 2) - (launchbarHeight / 2);
+    //Init to top-left
```
DivergentEuropeans: I may have mentioned this before, I don't remember... but we have a little bit of a smoothing bump to our windows for where they hit the edge of the browser, the real value is probably somewhere between 2-4px as the true start.
Here's a screenshot of what the difference is in the top left corner, it's a small thing but it's noticeable so we should be consistent with a bump or flat, and not both
image

Otherwise, everything is good!

jplinardon:
Sean to work with Aarohi to make a decision (go/no go)

DivergentEuropeans:
This is pretty old and we haven't had time to work on this without any complaints to how App windows get spawned

PR #107, opened 2019-05-22
App2App PostMessage callback for iframes

1000TurquoisePogs:
- Cleanup to work for sample-iframe-app
- Bugfix to not just throw exceptions when the platform periodically echos to iframes (for testing purposes...)
- New functionality to attempt handling of double-iframe case seen in explorers & jupyter app... In order for this to work reliably, a reserved name is needed. Seems we're gravitating towards "zluxIframe" at this time, so we should cement that.

Other changes:
- Corejs recommended an update, I see no issues there
- html2canvas too, but also found out how to turn off its aggressive logging

REQUIRES https://github.com/zowe/zlux-platform/pull/24

### DIFF
--- a/virtual-desktop/src/app/application-manager/application-manager.service.ts
+++ b/virtual-desktop/src/app/application-manager/application-manager.service.ts
@@ -79,6 +79,11 @@ export class ApplicationManager implements MVDHosting.ApplicationManagerInterfac
        if (applicationInstance){
          let theIframe:HTMLElement|null = document.getElementById(`${IFRAME_NAME_PREFIX}${instanceId}`);
          if (theIframe){
+           this.logger.debug(`Checking if double-iframe`);
+           let secondIframe = (theIframe as HTMLIFrameElement).contentWindow.document.getElementById('zluxIframe');
+           if (secondIframe) {
+             theIframe = secondIframe;
+           }
            this.logger.debug(`PostMessage for instance=`,applicationInstance);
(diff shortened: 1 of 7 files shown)


virtual-desktop/src/app/application-manager/application-manager.service.ts
```diff
@@ -51,17 +51,45 @@ export class ApplicationManager implements MVDHosting.ApplicationManagerInterfac
...
     (window as any).ZoweZLUX.dispatcher.setPostMessageHandler( (instanceId:MVDHosting.InstanceId, message:any ) => {
        let applicationInstance:ApplicationInstance|undefined = this.applicationInstances.get(instanceId);
        if (applicationInstance){
-         this.logger.debug(`PostMessage for instance=${applicationInstance}, iframeID=${applicationInstance.iframeId}`);
-         let theIframe:HTMLElement|null = document.getElementById(applicationInstance.iframeId);
+         let theIframe:HTMLElement|null = document.getElementById(`${IFRAME_NAME_PREFIX}${instanceId}`);
          if (theIframe){
-           this.logger.debug(`PostMessage iframe found=`,theIframe);
+           this.logger.debug(`Checking if double-iframe`);
+           let secondIframe = (theIframe as HTMLIFrameElement).contentWindow.document.getElementById('zluxIframe');
```
1000TurquoisePogs: Leave a code comment to explain or make a constant about the fact that this is a reserved secret iframe keyword

PR #125, opened 2019-07-11
Prevent tab from switching Zowe application

timgerstel:
"If you have several zowe applications running and you hit tab enough times to reach the end of the first application, on the next tabs it will go into the next zowe application."
This behavior has been mitigated by adding keyup and keydown listeners for "tab" events. The viewport ID for the window which contains the active element is derived through the parent node tree, and is compared to the viewport ID of the currently focused window.

### DIFF
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
@@ -97,6 +97,42 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
         .filter(win => win.windowState.stateType === DesktopWindowStateType.Maximized)
         .forEach(win => this.refreshMaximizedWindowSize(win));
     });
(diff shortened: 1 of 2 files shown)


virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -97,6 +97,39 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
...
+    }
+    document.addEventListener('keyup', tabHandler, false);
+    document.addEventListener('keydown', tabHandler, false);
+  }
+
+  private getViewportIdFromDOM(element: any): Number{
+    var parentViewportElement: any;
+    var head = element;
+    while(head.parentNode !== document){
+      if(head.parentNode.nodeName.toLowerCase() === 'com-rs-mvd-viewport'){
```
DivergentEuropeans: Just adding comments from what we discussed for sake of memory:

'com-rs-mvd-viewport' should be a variable name, in case we change it down the road.

PR #131, opened 2019-07-30
Some snapshot optimization

1000TurquoisePogs:
Reorganized snapshot logic to not poll, and not snapshot when only 1 instance open per window, since the snapshots cannot be visible at this point.
Also changed to take a snapshot when the window goes out of focus, instead of when it goes into focus, as this gives a better picture of where it was at when you last looked at it.
The difference is visible in idle performance:

image


1000TurquoisePogs:
Actually performance in prior code gets worse with quantity of apps open, due to stacked polling. New one does not:
image

PR #134, opened 2019-08-05
Removed window header offset at the top

DivergentEuropeans:
Removes the offset in the window header towards the top (bug visible where you cannot drag window if you drag from the tippy top)

image


1000TurquoisePogs:
No regression seen, but this does enhance drag-ability

PR #135, opened 2019-08-06
Add context menu entry to open app standalone

jordanCain:
Signed-off-by: Jordan Cain <jordan.cain1@uk.ibm.com>


virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.ts
```diff
@@ -181,12 +181,23 @@ export class LaunchbarComponent {
       this.applicationManager.spawnApplication(this.propertyWindowPluginDef,this.getAppPropertyInformation(plugin));
     }  
   }
+
+  openStandalone(item: LaunchbarItem): void {
+    const plginType:string = item.plugin.getFramework();
+    if (plginType === 'iframe') {
+      // Still allows IFrames to comprehend URL parameters if address is copy/pasted later
+      window.open(`${location.origin}/ZLUX/plugins/${item.plugin.basePlugin.getIdentifier()}/web/`);
```
DivergentEuropeans: We'd prefer to have this not hard coded in case our URI changes (like it did from /bootstrap/... last year) but this doesn't happen often and I can't think of a better way to do it atm so this is probably alright

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.ts
```diff
@@ -181,12 +181,23 @@ export class LaunchbarComponent {
...
+      window.open(`${location.href}?pluginId=${item.plugin.basePlugin.getIdentifier()}`);
+    }
+  }
   
   onRightClick(event: MouseEvent, item: LaunchbarItem): boolean {
     var menuItems: ContextMenuItem[];
     if (item.instanceCount == 1) {
         menuItems = [
           { "text": this.translation.translate("Open New"), "action": ()=> this.openWindow(item)},
+          { "text" : "Open Standalone", "action": () => this.openStandalone(item)},
```
DivergentEuropeans: Hey this functionally works, which is great, but disrupts internationalization

image

So let's touch base on Slack when you see this, change it to accept translations (it won't break, if there's no translations present, it will just use the default supplied text string) and I'll touch base with our translators to give you the strings that you can add to the i18n assets

1000TurquoisePogs: Let's get some translations into this PR as your follow-up.

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.ts
1000TurquoisePogs: Can you try window.ZoweZLUX.uriBroker.pluginResourceUri(item.plugin, '') ??
https://github.com/zowe/zlux-app-manager/blob/master/bootstrap/src/uri/mvd-uri.ts#L105

jordanCain:
@DivergentEuropeans  I am seeing the same behaviour, the structure of the iFrame app means it doesn't have an index.html in the web directory which seems to be the issue. If I change the behaviour so that all apps are opened in the form `https://localhost:8544/ZLUX/plugins/org.zowe.zlux.bootstrap/web/?pluginId=plugin.id` it seems to work but then Iframe apps may fail that require url parameters for functionality.

As for the VT terminal it also doesn't work, both terminal emulators are angular framework apps.

DivergentEuropeans:
@jordanCain If I remember correctly, both apps are Iframes wrapped in an Angular framework. So basically an Angular app with an Iframe template body and a few extra goodies. But yeah, those should work then. Not sure why they don't, they have a main TS component.

jordanCain:
I spoke with @1000TurquoisePogs about the VT terminal not working standalone and he's aware of the problem and has a solution in mind, something to do with a parameter that should be optional but is currently mandatory preventing it from loading.

jordanCain:
@1000TurquoisePogs Not sure why this got closed?

1000TurquoisePogs:
I'm not sure either. I think github messed up somehow.
image

PR #140, opened 2019-08-23
Launchbar fixes (+properties app fixes)

1000TurquoisePogs:
Fixed the following:
open-new was missing from within start menu
start menu didn't restore a minimized app when clicking the app icon when instance already open. 
add id to properties app, so you can refer to IDs more easily. 
clicking properties in right click menu does not focus properties app. 
cant re-use properties app (try to open properties for 2 different apps - doesnt do anything)


1000TurquoisePogs:
The bulk of this diff is that I moved the right click menu behavior for icons out into a common file for both menu & launchbar.

PR #152, opened 2019-09-13
Separated tslint configuration files for base, react and angular.

code-blooded-developer:
The base tslint.json was having some angular rules which were not having the implementation as well. So added codelyzer for angular which implements the angular linting rules and tslint-react for react linting rules.


virtual-desktop/plugin-config/tslint.base.linient.json
```diff
@@ -0,0 +1,168 @@
...
+    "arrow-parens": true,
+    "arrow-return-shorthand": true,
+    "ban-types": {
+      "options": [
+        ["Object", "Avoid using the `Object` type. Did you mean `object`?"],
+        [
+          "Function",
+          "Avoid using the `Function` type. Prefer a specific function type, like `() => void`."
+        ],
+        ["Boolean", "Avoid using the `Boolean` type. Did you mean `boolean`?"],
```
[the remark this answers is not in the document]
DivergentEuropeans: I noticed you named this file a whole different lint file so it's not that important I suppose. But we also shouldn't commit it unless it's useful for other people. Also it's "lenient", slight misspelling : )

virtual-desktop/plugin-config/tslint.base.linient.json
```diff
@@ -0,0 +1,168 @@
...
+      "options": ["spaces"]
+    },
+    "interface-name": {
+      "options": ["always-prefix"]
+    },
+    "interface-over-type-literal": true,
+    "jsdoc-format": true,
+    "label-position": true,
+    "max-classes-per-file": { "options": 1 },
+    "max-line-length": { "options": 120 },
```
DivergentEuropeans: If this is supposed to be lenient, should there be a max line length at all? Maybe a "ignore-pattern", maybe something like this for the main TS things
```
"ignore-pattern": "^import [^,]+ from |^export | implements"
```
This will ignore long imports & class definitions. Something to consider if it's useful for you

DivergentEuropeans: Also, the base.json has 140 as max line length, so you should use 140 or more

virtual-desktop/plugin-config/tslint.react.base.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["./tslint.base.linient.json", "tslint:latest", "tslint-react"]
```
DivergentEuropeans: Also, need @1000TurquoisePogs 's input on whether or not these should go into the base JSON's for React and Angular. If any property in these files happens to be more restrictive than base, we risk an error to any existing app at compile time

I think these extends should go in "tslint.react.lenient.json" and "tslint.angular.lenient.json"

1000TurquoisePogs:
I tested this as working, but should wait for conflict resolution & responses to Lenny's comments which are valid points.

virtual-desktop/plugin-config/tslint.base.json
```diff
@@ -2,28 +2,16 @@
   "rules": {
     "callable-types": true,
     "class-name": true,
-    "comment-format": [
-      true,
-      "check-space"
-    ],
+    "comment-format": [true, "check-space"],
```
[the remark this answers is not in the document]
code-blooded-developer: Removed unnecessary changes.

virtual-desktop/plugin-config/tslint.react.base.json
code-blooded-developer: The idea is to replace tslint.base.json with tslint.base.lenient.json in the future because it has some rules specific to angular that I am separating out. Also, the base file should only have the recommended rules and if some rules are not applicable to importer they can override it. So I have include only the base rules.

code-blooded-developer:
I have resolved the conflicts and worked on Leanid's comments.

virtual-desktop/plugin-config/tslint.base.linient.json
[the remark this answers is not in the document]
DivergentEuropeans: Yes, I meant someone would _want_ to use the primitive type to create an object with similar attributes but add a property. I've never had to do this and I'm assuming it's not good practice so I have no good opinion

virtual-desktop/package.json
```diff
@@ -28,6 +28,7 @@
     "angular-l10n": "~5.2.0",
     "angular2-template-loader": "~0.6.2",
     "bootstrap": "~4.3.1",
+    "codelyzer": "^5.1.0",
```
DivergentEuropeans: You're using "^" here instead of "~" which is fine if that's what you intended to do, but just make sure that's what you want it to do

https://dev.to/saurabhdaware/but-what-the-hell-is-package-lock-json-b04

virtual-desktop/package.json
```diff
@@ -49,6 +50,7 @@
     "style-loader": "~0.21.0",
     "ts-loader": "~4.4.2",
     "tslint": "~5.10.0",
+    "tslint-react": "^4.0.0",
```
DivergentEuropeans: Same here

virtual-desktop/package.json
code-blooded-developer: Yeah, I realized this later and have changed them to tilde.

PR #162, opened 2019-09-26
standalone enhancements

1000TurquoisePogs:
Few enhancements for single-window-mode:
1) Utilize an App's icon as the favicon
2) Utilize an App's title (i18n too) as the window title
3) Allow the single window mode to display the standard zowe login, if desired. Otherwise, apps can continue to use their own login screens.
Resolves:
zowe/zlux#222

Related to:
zowe/zlux#200


virtual-desktop/src/app/window-manager/simple-window-manager/simple.component.ts
```diff
@@ -40,27 +41,35 @@ export class SimpleComponent implements OnInit {
     let requestedPluginID: string = (window as any)['GIZA_PLUGIN_TO_BE_LOADED'];
 
     if (!requestedPluginID) {
-      // Default back to sample plugin:
-      this.logger.warn("GIZA_PLUGIN_TO_BE_LOADED not set, defaulting to sample plugin!");
-      requestedPluginID = "com.rs.mvd.myplugin";
+      this.logger.severe("Plugin ID not given, cannot continue!");
```
DivergentEuropeans: If you removed the requestedPluginID from being set to a default, and you say you cannot continue, shouldn't the rest of the logic not run? requestPluginID is being used farther down the method. Either way if not breaking things, it seems to attempt to do pointless things

virtual-desktop/src/app/window-manager/simple-window-manager/simple.component.ts
```diff
@@ -40,27 +41,35 @@ export class SimpleComponent implements OnInit {
...
         };
         this.applicationManager.spawnApplication(plugin as DesktopPluginDefinitionImpl, launchMetadata);
         this.viewportId = this.windowManager.getViewportId(1);
       } else {
-        this.logger.warn("plugin to be loaded not found: "+pluginID);
+        this.logger.severe("Plugin to be loaded not found: "+pluginID);
       }
     })
-      .catch(x => this.logger.warn("plugin promise not returned"));
+      .catch(x => this.logger.severe("plugin promise not returned, cannot continue"));
```
DivergentEuropeans: Capitalize (for consistency's sake)

DivergentEuropeans:
@1000TurquoisePogs Additionally, I wanted to add that there are huge package-lock changes here. I am guessing all is as it should be, as you probably re-installed your dependencies and just committed the correct package-locks, but QA is heavily dependent on these and because there are package-lock changes for the whole desktop and its core desktop apps, it would be important to have QA re-run automated tests for the whole desktop to see that nothing broke.

PR #165, opened 2019-10-07
Feature/notification small bugfix

DivergentEuropeans:


virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-widget/launchbar-widget.component.css
```diff
@@ -250,6 +250,13 @@
   padding: 5px !important;
 }
 
+::ng-deep .org_zowe_zlux_editor_snackbar {
```
1000TurquoisePogs: Everything here is fine except this one shouldn't be in the framework code, and not a fan in general of theming the notification bar for the editor at this point in time.

PR #169, opened 2019-10-11
Added wrapping to Notification UI

DivergentEuropeans:
image

These dots programmatically appear when notification UI text body becomes too long.


sakeerthy:
CSS attribute that already does this, added in https://github.com/zowe/zlux-app-manager/pull/171

PR #173, opened 2019-10-24
Added check for testing purposes

sakeerthy:


1000TurquoisePogs:
This helps with determining if there was a complete fail or partial success, but may need more in the future with regards to if we were able to send to 100% of the clients requested. Regardless, this improves the info sent back about success.

PR #174, opened 2019-10-28
[WIP] iFrame adapter: added support for plugin defintion, logger, and launch metadata

timgerstel:
This pull request adds the following features to the iFrame adapter:
-Support for the following injectable angular2plugintokens: plugin definition, launch metadata
-Support for ZoweZLUX.Logger (global logger object)

Do not merge without: 
https://github.com/zowe/zlux-shared/pull/23
https://github.com/zowe/zlux-platform/pull/47
Signed-off-by: Timothy Gerstel <tgerstel@rocketsoftware.com>

### DIFF
--- a/bootstrap/web/iframe-adapter.js
+++ b/bootstrap/web/iframe-adapter.js
@@ -23,7 +23,10 @@ var numUnresolved = 0;
 var instanceId = -1;
 var parentLogger = new exports.Logger()
 if(parentLogger){
+    console.log('setting logger')
     parentLogger.addDestination(parentLogger.makeDefaultDestination(true, true, true))
+} else {
+    console.log('logger object did not exist.  Exports=', exports)
 }
 
 let messageHandler = function(message) {
--- a/bootstrap/web/iframe-adapter.js
+++ b/bootstrap/web/iframe-adapter.js
@@ -30,8 +30,9 @@ let messageHandler = function(message) {
         if(instanceId === -1 && data.dispatchData.instanceId !== undefined){
             instanceId = data.dispatchData.instanceId;
             translateFunction('registerAdapterInstance', []);
-        } else if(instanceId !== -1 && data.dispatchData.instanceId !== instanceId){
-            console.warn('Desktop attempted to change instanceId for iframe instance=', instanceId);
+        } else if(instanceId !== -1 && data.dispatchData.instanceId !== undefined 
+                    && data.dispatchData.instanceId !== instanceId){
+            console.warn('Desktop attempted to change instanceId for iframe instance=', instanceId, 'message=', message);
         }
     }
     if(data.key === undefined || data.instanceId != instanceId) return;
--- a/bootstrap/web/iframe-adapter.js
+++ b/bootstrap/web/iframe-adapter.js
@@ -21,10 +21,34 @@ var closeHandlers = {};
 var curResponseKey = 0;
 var numUnresolved = 0;
 var instanceId = -1;
-var parentLogger = new exports.Logger()
-if(parentLogger){
-    parentLogger.addDestination(parentLogger.makeDefaultDestination(true, true, true))
+var DEFAULT_GET_LOGGER_TIMEOUT = 3000;
+var parentLogger = null;
+
+// This function uses setTimeout due to the fact that the Logger class must be included in the html script, and therefore 
+// is not guaranteed to load before this script
+function initLogger(maxRetries, timeout){
+    if(typeof getLogger.retries === 'undefined'){
+        getLogger.retries = 0;
+    }
+    if(exports.Logger){
+        console.log('Initialized Logger')
+        parentLogger = new exports.Logger();
+        parentLogger.addDestination(parentLogger.makeDefaultDestination(true, true, true))
+        return;
+    }
+    setTimeout(() => {
+        if(getLogger.retries > maxRetries){
+            console.log('Unable to initialize Logger')
+            return;
+        }
+        if(exports === {} || exports.Logger === undefined){
+            getLogger.retries++;
+            getLogger(maxRetries, timeout);
+        }
+    }, timeout);
 }
+
+initLogger(5, DEFAULT_GET_LOGGER_TIMEOUT);
 
 let messageHandler = function(message) {
     if(message.data.dispatchData){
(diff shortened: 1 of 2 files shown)


virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin.component.ts
```diff
@@ -246,6 +251,14 @@ export class IFramePluginComponent {
             this.responses[args[0]].resolve();
           }
           return undefined;
+        case 'getPluginDefinition':
+          try{
+            return JSON.parse(JSON.stringify(this.pluginDefintion))
+          }catch (e){
+            return undefined;
+          }
+        case 'getLaunchMetadata':
```
1000TurquoisePogs: Don't pull, since the angular & react API is a push: when the constructor of the main component executes, the injectables are present whether or not they are used. Make the iframe API the same way in that when the iframe says it is ready, the desktop can post a message that contains all info necessary to create objects that are like those injectables. So, you need the iframe adapter to have some sort of event that the iframe code listens on to await this constructor response.

iframe --- READY ---> desktop
desktop ---- pluginDef, logger ID, ways to listen on window&viewport ----> iframe adapter
iframe adapter ---- event containing pluginDef, actual component logger, viewport & window objects ---> iframe code

virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin.component.ts
```diff
@@ -56,6 +56,25 @@ export class IFramePluginComponent {
...
+        this.frameSource.postMessage({
+          key: -1,
+          constructorData: {
+            pluginDef: JSON.parse(JSON.stringify(this.pluginDefintion)),
+            launchMetadata: this.launchMetadata
+          },
+          instanceId: this.instanceId
+        }, '*')
+      }catch(e){
+        this.frameSource.postMessage({
```
1000TurquoisePogs: This must have a log message, so that we can spot if this happens, because it should never happen.

bootstrap/web/iframe-adapter.js
```diff
@@ -14,24 +14,29 @@
...
 var numUnresolved = 0;
 var instanceId = -1;
 
 let messageHandler = function(message) {
-    if(message.data.dispatchData){
-        if(instanceId === -1 && message.data.dispatchData.instanceId !== undefined){
-            instanceId = message.data.dispatchData.instanceId;
+    let data = message.data;
+    if(data.dispatchData){
+        if(instanceId === -1 && data.dispatchData.instanceId !== undefined){
```
1000TurquoisePogs: This should have an else{} which logs a warning, because it would be very weird if the desktop tried to tell you that your instance ID was changing.

bootstrap/web/iframe-adapter.js
```diff
@@ -296,6 +299,10 @@ var ZoweZLUX = {
     }
 }
 
+ZoweZLUX.logger = new exports.Logger();
+ZoweZLUX.logger.addDestination(ZoweZLUX.logger.makeDefaultDestination(true, true, true))
```
1000TurquoisePogs: You need some more true here. `logger.addDestination(logger.makeDefaultDestination(true,true,true,true,true));`

virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin.component.ts
```diff
@@ -72,8 +72,10 @@ export class IFramePluginComponent {
             pluginDef: {},
             launchMetadata: this.launchMetadata
           },
-          instanceId: this.instanceId
-        }, '*')
+          instanceId: this.instanceId,
+          error: 'Unable to parse plugin definition'
+        }, '*');
+        console.warn('Unable to parse plugin defintion');
```
1000TurquoisePogs: This needs to be a logger owned by the desktop.
Give this a try:
import { BaseLogger } from 'virtual-desktop-logger';
then just set this.logger to BaseLogger.

bootstrap/web/iframe-adapter.js
```diff
@@ -69,7 +76,7 @@ let messageHandler = function(message) {
                 console.log('restored')
```
DivergentEuropeans: This is bootstrap JS context, so I believe a logger is importable here right?

bootstrap/web/iframe-adapter.js
```diff
@@ -69,7 +76,7 @@ let messageHandler = function(message) {
...
                 return;
             case 'windowEvents.moved':
-                console.log('moved')
+                //console.log('moved')
```
DivergentEuropeans: This is bootstrap JS context, so I believe a logger is importable here right?

bootstrap/web/iframe-adapter.js
```diff
@@ -86,6 +93,8 @@ let messageHandler = function(message) {
     }
 }
 
+window.addEventListener('message', messageHandler);
+
 window.addEventListener("load", function () {
     console.log('iFrame Adapter has loaded!');
```
DivergentEuropeans: Same here

virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin.component.ts
DivergentEuropeans: @1000TurquoisePogs  There's a log message a few lines down, but because this error is caught, I don't think it will get outputted to the console, so it would be good to add that 'e' in

virtual-desktop/src/app/plugin-manager/plugin-factory/iframe/iframe-plugin.component.ts
```diff
@@ -113,18 +138,22 @@ export class IFramePluginComponent {
...
+            key: key,
+            originCall: 'windowActions.spawnContextMenu',
+            instanceId: this.instanceId,
+            contextMenuItemIndex: i
+          }, '*')
+        }
       }
+      return copy;
+    }catch(e){
+      return undefined;
```
DivergentEuropeans: Would it be useful to log the error here too? I'm not familiar with this code base, it's your call

1000TurquoisePogs: Yes, let's put a log here stating what error occurred.

PR #176, opened 2019-11-11
desktop keybinding initial commit

NakulManchanda:
Added desktop level keyup event, filtered on CTRL+ALT  for now, registered on node **rs-com-mvd-desktop**

Also, added sample event CTRL+ALT+M to toggle desktop menubar


NakulManchanda:
i think it would be a good starter to add more events - added first keybinding to demo usage of new **KeybindingService** at desktop level

NakulManchanda:
updated one minor bug, added dco and added two more sample events

CTRL+ALT+DOWN_ARROW - minimizeToggle
CTRL+ALT+UP_ARROW - maximizeToggle

1000TurquoisePogs:
This looks good, but I'd like to give it a test to see that we don't encounter focus issues. If we don't, then I think it would be good to merge and/or review the default keybindings.

1000TurquoisePogs:
I don't see the issues that Lenny described, but I do suspect 1 issue: no event.stopPropagation() is used. I think this means if an app also used these keybindings then both a minimize + some app action would occur.

NakulManchanda:
> I don't see the issues that Lenny described, but I do suspect 1 issue: no event.stopPropagation() is used. I think this means if an app also used these keybindings then both a minimize + some app action would occur.

added event.stopImmediatePropagation

NakulManchanda:
@1000TurquoisePogs  this works now as expected, thanks for your suggestion

Changes:
Registering keyup at document level - as this works better - other nodes loose focus
stopping even propagation of all ctrl+alt event combination - let me know if you don't like it
capture true - now even if any app try to define same shortcut, desktop get it first, also stop propagation now


Next Steps: IN next PR
1) Keyboard navigation of start menu
2) switch windows navigation

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/keybinding.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.keyUpHandler = this.keyUpHandler.bind(this);
+  }
+
+  registerKeyUpEvent() {
+    document.addEventListener('keyup', this.keyUpHandler, true);
+  }
+
+  keyUpHandler(event: KeyboardEvent) {
+    if(event.altKey) {
+      console.log('desktop' + event.which);
```
DivergentEuropeans: We should probably use the zowe logger here (if the log is still necessary)

struga0258:
All the apps minimize and maximize for me and it seems to be working fine.

DivergentEuropeans:
> All the apps minimize and maximize for me and it seems to be working fine.

James and I are looking into what is happening. I tried James' deployed version of Zowe, that works for him, and it doesn't work on my computer. Our browsers are identical versions of Chrome, very very strange...

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/keybinding.service.ts
NakulManchanda: fixing

NakulManchanda:
1) worked for me too for iframe apps tried for jes-explorer, 
2) alt+ctrl is key that was removed while preparing for testing
3) some camel case correction

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/keybinding.service.ts
NakulManchanda: fixed

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -121,6 +125,21 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
     }
     document.addEventListener('keyup', tabHandler, false);
     document.addEventListener('keydown', tabHandler, false);
+
+    this.appKeyboard.registerKeyUpEvent();
+    this.appKeyboard.keyUpEvent
+      .subscribe((event:KeyboardEvent) => {
+        if (event.which === KeyCode.DOWN_ARROW) {
+          if(this.focusedWindow) {
+            this.minimizeToggle(this.focusedWindow.windowId);
```
NakulManchanda: minimize code

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -121,6 +125,21 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
...
+          }
+        }
+        else if (event.which === KeyCode.UP_ARROW) {
+          if(this.focusedWindow) {
+            this.maximizeToggle(this.focusedWindow.windowId);
```
NakulManchanda: maximize code

DivergentEuropeans:
> 1. worked for me too for iframe apps tried for jes-explorer,
> 2. alt+ctrl is key that was removed while preparing for testing
> 3. some camel case correction
> 
> i refreshed all my code yesterday may be try that @DivergentEuropeans
> 
> can you try CTRL+ALT+M - to toggle menu?
> because maximize/minimize is not much different, i use all the existing code for them
> but can check if i find something, but unable to replicate the issue
> see code if you think something is missing.
> 
> may be not hooking at right spot - either too deep or there should be some other event which should trigger.

Hey Nakul, it has nothing to do with the code, as I tried James' deployed version of Zowe. It simply did not work when viewed on my computer. Yes, all the other keybindings work for me, including maximize and minimize, just those specific ones didn't work for Iframe apps. Unfortunately, I can't replicate the issue outside my machine. EITHER WAY, as I said, this isn't an issue for this PR and can be added as some extraneous/random bug fix later. I see you fixed the log so all looks good to me

NakulManchanda:
@DivergentEuropeans @1000TurquoisePogs 
i am able to confirm when i click inside sample iframe app, it able to steal focus in a way that keyup handler at top level parent document didnt trigger even with capture=true, even ctril+atl+M menu toggle didnt work

We can try extending it in next PR if able to find solution, thanks!

PR #178, opened 2019-11-20
Added a check for standalone app mode for iframes ONLY

DivergentEuropeans:
- Allows for things like this:
image


1000TurquoisePogs:
Seems to work but would like to see things scoped into ZoweZLUX.
ZoweZLUX.iframe = {pluginDef, launchContext...}
ZoweZLUX._iframeAdapter = {these, otherthingsyoushouldntsee...}

PR #187, opened 2020-02-26
uri broker fix for apiml use

1000TurquoisePogs:
Bugfix for when using editor under apiml: apiml does not like encoded slashes


1000TurquoisePogs:
The build failed due to timeout but is succeeding here: https://wash.zowe.org:8443/blue/organizations/jenkins/zlux-core/detail/zlux-core/1793/pipeline

PR #190, opened 2020-03-05
Add lang hint to html head based on user preference as soon as known

1000TurquoisePogs:
This could have detrimental effects if the apps do not support the language and do not specify otherwise.
Try testing the editor versus the sample angular app to compare how japanese operates, for example.

Made for testing with https://github.com/zowe/zlux/issues/394


1000TurquoisePogs:
Tested within 394

PR #195, opened 2020-03-13
Feature/app generator

1000TurquoisePogs:
Adding the app generator, shown at SHARE, after cleanup.


system-apps/app-generator/webClient/src/app/app.component.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  createCallback() {
+    var t = this;
+    return function(plugin) {
+      t.appList.push(plugin);
+    }
+  }
+
+	ngOnInit() {
+    const MY_PLUGIN_ID = ZoweZLUX.iframe.pluginDef.basePlugin.identifier;
```
lchudinov: I get an error on this line in Chrome console: `Cannot read property 'basePlugin' of undefined`. Note: that I had to build app-generator in dev mode to get the meaningful message.

1000TurquoisePogs: This means there is a timing issue. Needs a fix.

Joe-Winchester:
@1000TurquoisePogs - is the app generator in a build, or is there a way for someone to download/install/test it ?  I know a few folks who want to demo it to customers/ISVs.

DivergentEuropeans:
> @1000TurquoisePogs - is the app generator in a build, or is there a way for someone to download/install/test it ? I know a few folks who want to demo it to customers/ISVs.

Hey @Joe-Winchester ,
Not sure if this answers your question! But from a quick glance, it looks like the App Generator has a server component via https://github.com/zowe/zlux-app-manager/pull/195/files#diff-2705d2e050a835c1662c06812d5515bf and a standard web component via https://github.com/zowe/zlux-app-manager/pull/195/files#diff-9424460866b29f05f50ff4439da010eb 
So that means you can download, install, & test it via the normal procedure of cloning the repository, installing & building it with npm commands, & then adding the JSON plugin identifiers within the /plugins/ location of the Zowe app server.

This is nearly identical to how one would set up the Sample Angular App https://github.com/zowe/sample-angular-app via its web (webClient) & server (nodeServer) components. I haven't tried this app out myself *yet* but if you're interested in doing this, feel free to reach out on Slack & I can try to help. Otherwise, if a custom build or something similar would be more suitable, @1000TurquoisePogs would be your guy (he's on vacation for a few bus. days) 🏎️

1000TurquoisePogs:
I believe this is now fixed (moved the timing to a later function). If it is not, I will fix it in a follow-up PR.

PR #207, opened 2020-03-31
Emit an event when login screen opens

DivergentEuropeans:


DivergentEuropeans:
Will be covered by other work

PR #210, opened 2020-04-02
Session Events

sakeerthy:
Added event emitter for session changes, login, logout, sessionExpire for Angular, React, and iFrame applications


bootstrap/web/iframe-adapter.js
```diff
@@ -59,23 +59,35 @@ let messageHandler = function(message) {
             //The following cases should be implemented by the user,
             //therefore users will need an eventListener for "message" events
             case 'windowEvents.minimized':
-                console.log('minimized')
+                let minimized = new CustomEvent('windowEvents', {detail: {event:'minimized'}});
```
1000TurquoisePogs, line 62: all of these custom events should start with ZoweZLUX, so that they dont have a chance of conflicting with whatever is also in the iframe

1000TurquoisePogs: For example, ZoweZLUX.windowEvents, ZoweZLUX.sessionEvents

virtual-desktop/src/app/window-manager/simple-window-manager/simple-window-manager.service.ts
```diff
@@ -24,6 +25,7 @@ export class SimpleWindowManagerService implements MVDWindowManagement.WindowMan
   readonly windowResized: EventEmitter<{ width: number, height: number }>;
   private applicationManager: MVDHosting.ApplicationManagerInterface;
   private pluginDef: MVDHosting.DesktopPluginDefinition;
+  private authenticationManager: any;
```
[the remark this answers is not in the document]
1000TurquoisePogs: It is possible that by altering the type info here that it could break plugins that currently use it.
I don't want to touch it at this time, please. I know this wont cause a runtime issue but it could cause a built time issue.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -70,6 +71,7 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
   private viewportManager: MVDHosting.ViewportManagerInterface;
   private pluginManager: MVDHosting.PluginManagerInterface;
   public screenshotRequestEmitter: Subject<{pluginId: string, windowId: MVDWindowManagement.WindowId}>; 
+  private authenticationManager: any;
```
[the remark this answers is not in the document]
1000TurquoisePogs: It is possible that by altering the type info here that it could break plugins that currently use it.
I don't want to touch it at this time, please. I know this wont cause a runtime issue but it could cause a built time issue.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -332,12 +335,38 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
     };
   }
 
+  private generateSessionEventsProvider(): Angular2PluginSessionEvents {
+    let loginSubject = new Subject<void>();
+    let logoutSubject = new Subject<void>();
+    let sessionExpireSubject = new Subject<void>();
+    this.authenticationManager.loginScreenVisibilityChanged.subscribe((eventReason: LoginScreenChangeReason) => {
```
1000TurquoisePogs, line 342: You should not have the subscribe here. Rather, this subscribe should be part of the window manager service constructor, and you store each of the subjects in a map of key=windowid, and value=the 3 subjects, then during a event that comes from the visibility change, you can run through the list and call the event on the subjects that are still valid.
However, upon closewindow or destroywindow ( i forgot which) you can clean up the map so that things dont get called.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -332,12 +335,38 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
     };
   }
 
+  private generateSessionEventsProvider(): Angular2PluginSessionEvents {
```
1000TurquoisePogs, line 338: pass in the window id here

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -332,12 +335,38 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
...
+    let loginSubject = new Subject<void>();
+    let logoutSubject = new Subject<void>();
+    let sessionExpireSubject = new Subject<void>();
```
1000TurquoisePogs, line 341: store these according to window id in a map

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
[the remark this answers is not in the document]
1000TurquoisePogs: Nevermind, Lenny's right, this can safely be not 'any'.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
DivergentEuropeans: @1000TurquoisePogs and I discussed, Each of the other events relate to the plugins at the plugin level, things like the window events & viewport events etc, but session events are for the session. It doesn't matter if you're in the Angular App or the Editor, a logout is still a logout so we could in theory have at max 3, but that would make it more difficult to control the events at lifecycle without tying them down to the window level. Do whatever you think is better @sakeerthy

PR #213, opened 2020-04-06
zowe desktop start menu keybindings

NakulManchanda:


NakulManchanda:
@1000TurquoisePogs  @aarohimshah1 
Added new shortcuts to navigate start menu with keyboard:
- CTRL+ALT+M - to open start menu
- focus switch to search bar when start menu opens, whether with mouse click or above keyboard shortcut
- then up and down key to navigate app list,
- then enter to open app
- with right arrow key, we can open right click menu, this choice has some conflict with cursor movement, but is not big problem in my opinion, open for suggestions though

Feel free to try and let me know if any feedback, thanks!

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
```diff
@@ -146,11 +170,108 @@ export class LaunchbarMenuComponent implements MVDHosting.LoginActionInterface{
...
+  private unMarkContextMenuPresent():void {
+    setTimeout(()=> {
+      console.log('toggle menu present');
+      this.isContextMenuPresent = false;
+    },1000)
+  }
+
+  @HostListener('keyup', ['$event'])
+  onKeyUp(event: KeyboardEvent) {
+    console.log('menu keyboard event',this.isContextMenuPresent);
```
DivergentEuropeans, line 190: Rogue console.log (or use zowe logger)

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
```diff
@@ -146,11 +170,108 @@ export class LaunchbarMenuComponent implements MVDHosting.LoginActionInterface{
...
+    }
+  }
+
+  private markContextMenuPresent():void {
+    this.isContextMenuPresent = true;
+  }
+
+  private unMarkContextMenuPresent():void {
+    setTimeout(()=> {
+      console.log('toggle menu present');
```
DivergentEuropeans, line 183: Rogue console.log (or use zowe logger)

virtual-desktop/src/app/context-menu/context-menu.component.ts
```diff
@@ -20,18 +20,21 @@ import {
...
   AfterViewInit,
-  QueryList
+  QueryList,
+  OnDestroy,
+  OnInit
 } from '@angular/core';
 
 import { ContextMenuItem } from 'pluginlib/inject-resources';
 import { DomSanitizer } from '@angular/platform-browser';
+import { ContextMenuService } from './contextmenu.service';
```
DivergentEuropeans, line 30: I don't think this is a good use of the service. Can you not just add a "showContextMenu" boolean, or something, to the ContextMenuComponent and then use that as a conditional for whether certain keyboard bindings are on or off. The Keybinding service shouldn't be "alerted" when it can use a key again, it should try to use that key and if a case doesn't match, nothing happens for that key press. Using a service is slow here (you can test this by pressing "Right" to open the menu, pressing "Escape" quickly to close, and trying to press "Right" again it won't work) And I understand that this was intentional due to your 1000ms timeout for your subscription, but that's why it's not a good idea. A boolean will also work here because you're never going to have multiple context menu's (or you shouldn't) so keeping track of multiple components isn't a problem here. Because the ContextMenu component won't exist when you don't have it spawned, your initial check might want to be like 
`if (contextMenu && contextMenu.showContextMenu)`

Seeing as right now, it seems contextMenu is only available when it's visible, you can probably get away with just doing the 1st check, but adding that boolean might be a cleaner, easier to see way of doing it

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
```diff
@@ -106,16 +119,27 @@ export class LaunchbarMenuComponent implements MVDHosting.LoginActionInterface{
   
   activeToggle(): void {
     this.isActive = !this.isActive;
+    // gain focus and clear on toggle when active
+    if(this.isActive) {
+      setTimeout(() => {
+        this.searchAppInputRef.nativeElement.focus();
```
DivergentEuropeans, line 125: Why is this wrapped in a setTimeout?

Also, there's an HTML way of giving an element default focus when it spawns. If I remember it's the "autofocus" attribute but not sure on performance of yours vs that one

NakulManchanda: 1) settimeout so element added to DOM otherwise nativeElement was null
2) autofocus just works one time for some reason and then it dont focus any more, tried that html attribute

virtual-desktop/src/app/context-menu/context-menu.component.ts
NakulManchanda: Sure, if you think it works i can try this and let you know if it does the job or not, ofcourse simpler the better

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
NakulManchanda: Will remove

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
NakulManchanda: Will remove

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
DivergentEuropeans: Ah, I see. Thanks, good to know @NakulManchanda

virtual-desktop/src/app/context-menu/context-menu.component.ts
DivergentEuropeans: Sure, that sounds good. Lmk if you need help

DivergentEuropeans: If it becomes too difficult to import the component, maybe due to module inconsistencies or something, it will probably be fine to use the service, just do it in a way without the 1000ms by doing isShown checks, instead of emitter events

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
NakulManchanda: removed console.log statement

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
NakulManchanda: removed function itself

virtual-desktop/src/app/context-menu/context-menu.component.ts
NakulManchanda: removed service altogether and removed 1000ms delay as well, found one way

NakulManchanda:
> 3\. Clicking on the top portion of the context menu UI means I can no longer use keys to
>     move around the apps

fixed bug, it was stealing focus out of search box, now i keep focus on search box, which has host listener attached to hear all keyboard events

NakulManchanda:
> 2\. Mouse hover should change app selection in context menu (or you can take the Windows approach and add a second different-colored selection for the mouse)

added mouse over, to change active selection, see if it looks fine to you

NakulManchanda:
> 1. "Left" key should dismiss context menu too if "Right" opens it
@DivergentEuropeans addressed all other feedback, except this one
If we want to add "left" for dismissing we have to make change ArrowLeft case of context menu and then regression context menu, i will suggest we can do this separately, to avoid regression issues. Esc key was new so it was easy to add.

NakulManchanda:
Summary, for this revision: @DivergentEuropeans 

iterated on this one, resolved all concern,
except adding Left key to dismiss context menu, which i suggest we can do separately

**Issues Addressed**
1) No context menu service  now
2) No 1000 ms delay
3) Retain focus on search on click on top area of menu bar
4) added mouse hover to change active selection
5) adjusted context menu location little bit too, so icon and app name are visible when we context menu with right arrow key

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
```diff
@@ -47,6 +50,8 @@ export class LaunchbarMenuComponent implements MVDHosting.LoginActionInterface{
   propertyWindowPluginDef : DesktopPluginDefinitionImpl;
   public authenticationManager : MVDHosting.AuthenticationManagerInterface;
   public appFilter:string="";
+  public activeIndex:number;  
+  private isContextMenuPresent:boolean = false;
```
DivergentEuropeans, line 54: One small thing, can you set this in the constructor with the other initializations?

NakulManchanda: done

PR #224, opened 2020-04-27
replace custom keepcursor with event.preventDefault

NakulManchanda:


1000TurquoisePogs:
simpler=better here

PR #234, opened 2020-06-03
Cross-launch via URL

lchudinov:
## Cross-launch via URL

This PR depends on https://github.com/zowe/zlux-platform/pull/59

Sample URL: `https://localhost:8544/ZLUX/plugins/org.zowe.zlux.bootstrap/web/?app2app=org.zowe.zlux.ng2desktop.webbrowser:launch:create:data:{"url":"https://github.com/zowe/zlux-app-manager/pull/234","enableProxy":true}&app2app=org.zowe.zlux.ng2desktop.webbrowser:message:create:data:{"url":"https://github.com/zowe/zlux-app-manager/pull/234","enableProxy":true}&app2app=org.zowe.zlux.ng2desktop.webbrowser:message:create:org.zowe.zlux.test.action:{"data": {"url":"https://github.com/zowe/zlux-app-manager/pull/234","enableProxy":true}}`
Query parameter format is `?app2app={pluginId}:{actionType}:{actionMode}:{formatter}:{contextData}&app2app={pluginId}:{actionType}:{actionMode}:{formatter}:{contextData}`
* `pluginId` - application identifier, e.g. `'org.zowe.zlux.ng2desktop.webbrowser'`
* `actionType` - `'launch' | 'message'`
* `actionMode` - `'create' | 'system'`
* `formatter` - `'data'` | actionId
* `contextData` - context data in form of JSON

### DIFF
--- a/virtual-desktop/src/app/start-url-manager/start-url-manager.service.ts
+++ b/virtual-desktop/src/app/start-url-manager/start-url-manager.service.ts
@@ -9,38 +9,19 @@
 */
 
 import { Injectable } from '@angular/core';
-import { ZluxErrorSeverity, ZluxPopupManagerService } from '@zlux/widgets';
-import { TranslationService } from 'angular-l10n';
 import { BaseLogger } from 'virtual-desktop-logger';
-import { PluginLaunchURLParams, TargetPluginMode, isValidTargetPluginMode } from './plugin-launch-url-params';
+import { App2AppArgs } from './app2app-args';
+import { App2AppArgsParser } from './app2app-args-parser.service';
 
-
-const TARGET_PLUGIN_ID = 'targetPluginId';
-const TARGET_PLUGIN_DATA = 'targetPluginData';
-const TARGET_PLUGIN_MODE = 'targetPluginMode';
-const TARGET_PLUGIN_INSTANCE_ID = 'targetPluginInstanceId';
-
-type ActionTargetMode = typeof ZoweZLUX.dispatcher.constants.ActionTargetMode;
 @Injectable()
 export class StartURLManager implements MVDHosting.LoginActionInterface {
   private done = false;
-  private readonly popupOptions: { blocking: boolean; buttons: any[]; };
   private readonly logger: ZLUX.ComponentLogger = BaseLogger;
-  private readonly modeMap: Record<TargetPluginMode, ActionTargetMode> = {
-    'create': ZoweZLUX.dispatcher.constants.ActionTargetMode.PluginCreate,
-    'findAnyOrCreate': ZoweZLUX.dispatcher.constants.ActionTargetMode.PluginFindAnyOrCreate,
-    'findUniqueOrCreate': ZoweZLUX.dispatcher.constants.ActionTargetMode.PluginFindUniqueOrCreate,
-    'system': ZoweZLUX.dispatcher.constants.ActionTargetMode.System,
-  };
 
   constructor(
-    private popupManager: ZluxPopupManagerService,
-    private translation: TranslationService,
+    public parser: App2AppArgsParser,
   ) {
-    this.popupOptions = {
-      blocking: true,
-      buttons: [this.translation.translate('Close')],
-    };
+    this.registerTestAction();
   }
 
   onLogin(_username: string, _plugins: ZLUX.Plugin[]): boolean {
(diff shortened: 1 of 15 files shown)


1000TurquoisePogs:
this needs more features in order to work for certain programs, and in order to work for multiple programs. It appears to only work for 1 plugin, to launch it, and with a simple data structure.
It doesn't look like it would work for messaging the desktop, since you need message+system, not launch+system. the app target modes are really only useful for "message", not "launch". I believe "message" does "launch" if none are open yet.

consider this:
```
?app2app=org.zowe.zlux.ng2desktop.webbrowser:launch:create:data:{"url":"https://github.com/zowe/zlux-app-manager/pull/234","enableProxy":true};org.zowe.zlux.ng2desktop.webbrowser:launch:create:data:{"url":"https://google.com/","enableProxy":true};org.zowe.complicatedprogram:launch:create:org.zowe.complicatedprogram.complicatedstruct:{"a":1, "b": {"c": 2, "d": 3}};org.zowe.desktop:message:system:org.zowe.desktop.settheme:{"size":1, "colors": {"toolbar": "blue", "window": "black"}}
```

`:` separates part of each app2app request.
`;` separates each app2app request.
a request is
plugin + action type + action target mode + argument formatter + context data
org.zowe.zlux.ng2desktop.webbrowser:launch:create:data:{"url":"https://github.com/zowe/zlux-app-manager/pull/234","enableProxy":true};

"data" in this example means what you already have: take the contents and put them inside of {"data": thecontents}

But, there are app2app Actions which have something mapped from a different structure, such as let's talk about the desktop theme

org.zowe.desktop:message:system:org.zowe.desktop.settheme:{"size":1, "colors": {"toolbar": "blue", "window": "black"}}

maybe org.zowe.desktop.settheme is:
{"size": {"launchbarsize": data.size, "windowsize": data.size}}
Here, because the input is:
{"size":1, "colors": {"toolbar": "blue", "window": "black"}}
It means that the result is actually
{"size": {"launchbarsize": 1, "windowsize": 1}}
And the color is discarded.

If the URL does contain an ID such as org.zowe.desktop.settheme instead of "data", you must lookup or fail. Lookup is: https://github.com/zowe/zlux-platform/blob/master/base/src/dispatcher/dispatcher.ts#L465

virtual-desktop/src/app/start-url-manager/start-url-manager.service.ts
```diff
@@ -0,0 +1,141 @@
...
+    const targetPluginId = args.pluginId;
+    const type = args.actionType === 'launch' ? ActionType.Launch : ActionType.Message;
+    const mode = args.actionMode === 'create' ? ActionTargetMode.PluginCreate : ActionTargetMode.System;
+    let contextData: any = JSON.parse(args.contextData);
+    let action: ZLUX.Action;
+    let argumentData: any;
+    if (args.formatter === 'data') {
+      const actionTitle = 'Launch app from URL';
+      const actionId = 'org.zowe.zlux.url.launch';
+      const argumentFormatter = { data: { op: 'deref', source: 'event', path: ['data'] } };
```
1000TurquoisePogs, line 93: I think we could bake this into zlux-platform dispatcher, so it makes it at startup. That way, we could call `dispatcher.getAbstractActionById('data');`
The reason why is because I am not sure what happens to the dispatcher if we do multiple `makeAction` calls for the same name

virtual-desktop/src/app/start-url-manager/start-url-manager.service.ts
```diff
@@ -0,0 +1,141 @@
...
+      action = dispatcher.makeAction(actionId, actionTitle, mode, type, args.pluginId, argumentFormatter);
+      argumentData = { data: contextData };
+    } else {
+      const abstractAction = dispatcher.getAbstractActionById(args.formatter);
+      action = <ZLUX.Action>{
+        ...abstractAction,
```
DivergentEuropeans, line 99: I don't remember the specifics, but old versions of Node (maybe 6) made it such that the spread operator doesn't work & wouldn't get compiled correctly into the JS, instead Object.assign was used

Don't remember if this is still an issue, it might have been an issue on Z for me, it's been nearly a year...

https://github.com/zowe/zlux-server-framework/commit/5abf39b0dd790b15cf1bf0ba6f533e30e1187792

DivergentEuropeans, line 99: I don't remember the specifics, but old versions of Node (maybe 6) made it such that the spread operator doesn't work & wouldn't get compiled correctly into the JS, instead Object.assign was used

Don't remember if this is still an issue, it might have been an issue on Z for me, it's been nearly a year...

https://github.com/zowe/zlux-server-framework/commit/5abf39b0dd790b15cf1bf0ba6f533e30e1187792

DivergentEuropeans: This is zlux-app-manager, where in multiple places currently ... is being used so it was probably only an issue for server core

virtual-desktop/src/app/start-url-manager/start-url-manager.service.ts
lchudinov: I took a look at `https://github.com/zowe/zlux-platform/blob/master/base/src/dispatcher/dispatcher.ts#L679` and deeper. `dispatcher.makeAction()` doesn't register the action that it has created.

PR #235, opened 2020-06-04
Remove usage of ngx-color

DivergentEuropeans:
- Removes usage of ngx-color (replaces with custom hue selection bar, lightness swatches bar, color palette)

image


virtual-desktop/src/app/window-manager/mvd-window-manager/personalization-panel/personalization/personalization.component.html
```diff
@@ -31,22 +31,61 @@
...
+      'personalization-circle': true}" (click)="clickCircle(13)"></canvas>
+    <canvas #circle14 width="40" height="40" [ngClass]="{ 'personalization-circle selected': selectedCircle == 14,
+      'personalization-circle': true}" (click)="clickCircle(14)"></canvas>
+    <canvas #circle15 width="40" height="40" [ngClass]="{ 'personalization-circle selected': selectedCircle == 15,
+      'personalization-circle': true}" (click)="clickCircle(15)"></canvas>
+    <canvas #circle16 width="40" height="40" [ngClass]="{ 'personalization-circle selected': selectedCircle == 16,
+      'personalization-circle': true}" (click)="clickCircle(16)"></canvas>
+    <canvas #circle17 width="40" height="40" [ngClass]="{ 'personalization-circle selected': selectedCircle == 17,
+      'personalization-circle': true}" (click)="clickCircle(17)"></canvas>
+    <canvas #circle18 width="40" height="40" [ngClass]="{ 'personalization-circle selected': selectedCircle == 18,
```
DivergentEuropeans, line 70: Setting width & height in CSS vs canvas element does not work quite as expected

virtual-desktop/src/app/window-manager/mvd-window-manager/personalization-panel/personalization/personalization.component.ts
```diff
@@ -25,30 +27,283 @@ import { Colors } from '../../shared/colors';
...
+    gradient.addColorStop(4 / 6, 'blue');
+    gradient.addColorStop(5 / 6, 'indigo');
+    gradient.addColorStop(1, 'violet');
+
+    // Fill
+    ctx.fillStyle = gradient;
+    ctx.fillRect(0, 0, 389, 14);
+
+    // Save imgData
+    this.sliderImgData = ctx.getImageData(0, 0, canvasElem.nativeElement.width, canvasElem.nativeElement.height);
```
DivergentEuropeans, line 297: We save imgData instead of using getImgData to do a lookup every-time because of performance

virtual-desktop/src/app/window-manager/mvd-window-manager/personalization-panel/personalization/personalization.component.html
```diff
@@ -31,22 +31,61 @@
...
-      [color]="selectedColor" 
-      (onChangeComplete)="colorSelected($event.color)" 
-      [width]="380" 
-      [colors]='["#f44336", "#e91e63", "#9c27b0", "#673ab7", "#3f51b5", "#2196f3", 
-      "#03a9f4", "#00bcd4", "#009688", "#4caf50", "#8bc34a", "#cddc39", "#ffeb3b", "#ffc107", 
-      "#ff9800", "#252628", "#6d7176", "#f3f4f4"]'
-    ></color-circle>
+  <div class="personalization-color-1"> <!-- Color circle group -->
+    <div class="personalization-text" style="margin-left: -15px;">Select color</div>
+    <canvas #circle1 width="40" height="40" [ngClass]="{ 'personalization-circle selected': selectedCircle == 1,
```
1000TurquoisePogs, line 36: why did you make a canvas for every single circle? that's overkill and probably not very cpu efficient.
could just be css: https://www.w3schools.com/howto/howto_css_circles.asp

virtual-desktop/src/app/window-manager/mvd-window-manager/personalization-panel/personalization/personalization.component.html
```diff
@@ -31,22 +31,61 @@
...
   </div>
-  <div class="personalization-color-2">
-    <color-slider 
-      [color]="selectedColor" 
-      (onChange)="colorPreview($event.color)"
-      (onChangeComplete)="colorSelected($event.color)"
-    ></color-slider>
+  <div class="personalization-color-2"> <!-- Hue slider -->
+    <div class="personalization-text">or hue.</div>
+    <canvas #slider1 width="389" height="14" (click)="clickSlider($event)"></canvas>
```
1000TurquoisePogs, line 75: this is the only canvas object i expected, since gradient is hard to do in css.

virtual-desktop/src/app/window-manager/mvd-window-manager/personalization-panel/personalization/personalization.component.html
DivergentEuropeans: Looks like most resources point to them being similar in performance, while DOM based rendering is described as being more memory intensive than the two surprisingly. CSS animations are also typically slower than canvas however text-based rendering suffers heavily with canvas. Both of those don't apply here though. Canvas consensus summed up well with
"
Fast. Really fast. Because an immediate mode system doesn't maintain its own model, your code is all that stands between you and the browser redrawing. The many layers of abstraction that slow operations down simply do not exist in the immediate mode world.
"
If you're okay, I'm okay
https://stackoverflow.com/questions/5882716/html5-canvas-vs-svg-vs-div
https://www.kirupa.com/html5/dom_vs_canvas.htm
https://www.reddit.com/r/html5/comments/2vnrvm/performance_of_the_canvas_vs_performance_of_the/

PR #242, opened 2020-06-15
login activity, storage event changes

NakulManchanda:
Signed-off-by: Nakul Manchanda <nakul.manchanda@ibm.com>

PR #247, opened 2020-06-25
session expire storage evnt,renew session on login

NakulManchanda:


NakulManchanda:
This PR:
- new session event in storage - just to communicate - expire and logout 
- on receiving session expire from other tab - all tabs should logout
- now on some new login, session is renewed rather than just refreshing last activity timestamp

NakulManchanda:
it has following changes
- On new login we were not refreshing login, just detecting activity, now in addition any new login we refresh token so old tab will not expire now earlier than new tab
- Session event Logout and Expire are sent to all tabs

this is to further enhance issue https://github.com/zowe/zlux/issues/403

NakulManchanda:
This PR is ready for review - extension of [last item](https://github.com/zowe/zlux-app-manager/pull/242) we added
added new event for session
one simple test is login multiple windows and do logout on one, other open windows logout too
for now just triggering session expire and logout to other windows,
difference is session expire preserve open windows

DivergentEuropeans:
If I were to make one small suggestion, that would be to use removeItem to clean up the ZoweZLUX key, value pairs that are relevant like times & username. LocalStorage expires indefinitely, as far as I remember, so unless the user empties their cache, it will stay in the browser forever.

So kind of far-fetched, but a social engineering scenario could take place: you could find out when a user was last active & when they logged out by having access to their browser

NakulManchanda:
> If I were to make one small suggestion, that would be to use removeItem to clean up the ZoweZLUX key
**Here is an attempt to manage storage event subscription**

@DivergentEuropeans @1000TurquoisePogs 

for Testing:

Make these two changes
**const ACTIVITY_IDLE_TIMEOUT_MS = 30000;** - [login.component.ts](https://github.com/zowe/zlux-app-manager/blob/8d22aa78bc6461358ebcbcaff58a0c1dc75607e6/virtual-desktop/src/app/authentication-manager/login/login.component.ts#L22)

**const DEFAULT_EXPIRATION_MS = 36000;** - [zssAuth.js](https://github.com/zowe/zss-auth/blob/89350c2d52c859079b8cd7cdee58e09e44de9edb/lib/zssAuth.js#L19)

1 tab -
         login > logout or session expire,
         login again>  logout or session expire
2 tab
         login > logout or session expire,
         login again>  logout or session expire
3 tab
        login > logout or session expire,
        login again>  logout or session expire

Let me know how it looks, if we need to improve on something, or if it making it worse.

NakulManchanda:
- Register to storage event post-login (everywhere we had `setSessionTimeoutWatcher` )
- Un-register storage event on logout 
- manually  `clearTimeout(this.expirationWarning);` on logout as well
- removed storage key `lastActive`, `sessionEvent`

NakulManchanda:
image

NakulManchanda:
Session Expired:

Sender Tab:
image

Listener Tab:
image

PR #250, opened 2020-07-02
Create Auto Save Event

miteshgoplani:
Added AutoSave Event for Angular, React and Iframe Apps

- [x] Angular Example: https://github.com/zowe/sample-angular-app/pull/43
- [x] React Example: https://github.com/zowe/sample-react-app/pull/30

Dependent on: https://github.com/zowe/zlux-platform/pull/62/files

Signed-off-by: miteshgoplani miteshgoplani@gmail.com


virtual-desktop/src/app/window-manager/simple-window-manager/simple-window-manager.service.ts
```diff
@@ -93,6 +93,7 @@ export class SimpleWindowManagerService implements MVDWindowManagement.WindowMan
   private generateSessionEventsProvider(): Angular2PluginSessionEvents {
     const loginSubject = new Subject<void>();
     const sessionExpireSubject = new Subject<void>();
+    const autosaveEmitter = new Subject<any>();
```
1000TurquoisePogs, line 96: Because this is never used, it means single app mode will not have autosave. I think it should, and that can be done next. In single app mode, you would restore data from index 0 of the app requested. All other apps and all other instances of that app would be ignored.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -58,6 +59,7 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
   private _lastScreenshotPluginId: string = '';  
   private _lastScreenshotWindowId: number = -1;
   public showPersonalizationPanel: boolean = false;
+  private autoSaveInterval : number = 5000;
```
1000TurquoisePogs, line 62: This is good for testing but probably too frequent otherwise. 30 seconds, 1 minute, or 5 minutes seems like better defaults.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
```diff
@@ -429,12 +432,32 @@ export class WindowManagerService implements MVDWindowManagement.WindowManagerSe
     };
   }
 
+  private savePluginData(plugin:ZLUX.Plugin,windowId:number,data:any){
+    let pathToSave : any = 'pluginData' + '/' + 'app'
+    let fileNameToSave : string = plugin.getIdentifier() + '-' + windowId
```
1000TurquoisePogs, line 437: windowId makes the filename unique, but that may become a problem on restoration. 
You will get a few files,
pluginA-1
pluginA-6
pluginA-8

When you go to restore, you restore them all, but now they are 0, 1, 2. So, then you may see
pluginA-0
pluginA-1
pluginA-2

That's also OK, but what would not be OK would be:
pluginA-0
pluginA-1
pluginA-2
pluginA-6
pluginA-8

Because then you went from 3 to 5, which is not what the user saw.

There could be 2 solutions:

1. After successful load of the autosave data, mark a boolean so that the first time after load that a new autosave will occur, it will first delete the old autosave data. You can use this.http.delete to delete content. In this way, you will prevent going from 3 to 5 in the above example.

2. Instead of saving 1 file per plugin instance to the config service, just save 1 file for all, where the file is an array of the data from the instances. In this way, you do not need to worry about cleanup of old files. It will always be replacing the old data with the new data.

In 2, you need to do a GET of the file at startup, and then all future autosave events just overwrite the file.
The file can look like
```
{ instances: 
[ 
  { data1: "abc", data2: false, data3: -3 },
  { data1: "def", data2: true, data3: -3 }
  { data1: "xzc", data2: false, data3: 0 }
]
}
```

In this way, it does not matter what the window Id was, and the order is also unimportant, because this file would be constantly overwritten. All that matters is that there were 3 instances of the plugin, not the order or the unique number.

virtual-desktop/src/app/authentication-manager/authentication-manager.service.ts
```diff
@@ -178,6 +183,20 @@ export class AuthenticationManager {
       }).catch((err:any)=> {
         observer.error(err);
       });
+
+      this.HTTP.get<any>(ZoweZLUX.uriBroker.pluginConfigUri(ZoweZLUX.pluginManager.getDesktopPlugin(),'pluginData/app', undefined)).subscribe(res => {
```
1000TurquoisePogs, line 187: rather than `undefined`, if you give it a specific file name, you can get back an array like I have in the below comment.

virtual-desktop/src/app/application-manager/application-manager.service.ts
```diff
@@ -400,6 +400,11 @@ export class ApplicationManager implements MVDHosting.ApplicationManagerInterfac
   killApplication(plugin:ZLUX.Plugin, appId:MVDHosting.InstanceId):void {
     ZoweZLUX.dispatcher.deregisterPluginInstance(plugin,
                                                 appId);   // instanceId is proxy handle to isntance 
+    let filePath : any = 'pluginData' + '/' + 'app'
+    let fileNameToDelete : string = plugin.getIdentifier() + '-' + appId
+    this.http.delete(ZoweZLUX.uriBroker.pluginConfigUri(ZoweZLUX.pluginManager.getDesktopPlugin(),filePath,fileNameToDelete)).subscribe(()=>
```
1000TurquoisePogs, line 405: From my comment about using a single file (option 2), then in the case that you close 1 instance, the simplest way to update the autosave data is to send the autosave event to all remaining instances of that plugin, so that you can save the result as a smaller array.

an alternative is to get the current file and remove the index for the instance you just closed. in that way, you would need to have a way to track which index this was, like with a map or some mathematical relationship between the id and the index.

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
1000TurquoisePogs: nevermind, it appears you already did option 1 which is fine.

virtual-desktop/src/app/application-manager/application-manager.service.ts
miteshgoplani: Using Option 1

virtual-desktop/src/app/authentication-manager/authentication-manager.service.ts
miteshgoplani: Using Option 1

virtual-desktop/src/app/window-manager/mvd-window-manager/shared/window-manager.service.ts
miteshgoplani: Changed this to 60000

virtual-desktop/src/app/window-manager/simple-window-manager/simple-window-manager.service.ts
miteshgoplani: Done AutoSaving for Single App Mode

1000TurquoisePogs:
I noticed a performance problem: When you close a window, DELETE is called. But, if the window never had autosave data, then the DELETE is not needed. It would be good to have a simple map of each app and whether the app saved data. We could even add this property to the pluginDefinition.json of apps that support the feature
webContent: {
autosave: true
}

Because, if its not supported, then it is not useful to make the server do extra work.

miteshgoplani:
> I noticed a performance problem: When you close a window, DELETE is called. But, if the window never had autosave data, then the DELETE is not needed. It would be good to have a simple map of each app and whether the app saved data. We could even add this property to the pluginDefinition.json of apps that support the feature
> webContent: {
> autosave: true
> }
> 
> Because, if its not supported, then it is not useful to make the server do extra work.

@1000TurquoisePogs 
1. Resolved this by adding `autosave: true` to `pluginDefinition`
2. Updated the Sample Apps PRs

PR #265, opened 2020-07-27
Add support for iframes that reference apiml directly

1000TurquoisePogs:
If an iframe is for something on the apiml, then when the desktop is being accessed through apiml, the route could start with `/` so that the iframe app does not need hardcoded knowledge of the api gateway host & port.
This would also work equally well for specification of explicit paths to the app server.
Care must be taken to distinguish those two, as both abilities cannot be used at once.
In the future, we can expand upon this with env-var-like statements in the string to hint at the desired destination.

now depends upon https://github.com/zowe/zlux-platform/pull/63

### DIFF
--- a/bootstrap/package-lock.json
+++ b/bootstrap/package-lock.json
@@ -6,29 +6,35 @@
   "dependencies": {
     "@types/core-js": {
       "version": "2.5.3",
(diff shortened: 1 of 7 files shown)


1000TurquoisePogs:
This was done in a similar way with the destination parameter like so https://github.com/zowe/explorer-jes/blob/master/pluginDefinition.prod.json#L8

PR #286, opened 2020-09-25
App-generator features

jpreagan:
* Updated app-generator to use Carbon components
* Implemented input validation for Identifier and 'Create' button
* Added dropdown for licenses and license description box
* Changed framework to selectable tiles
* Changed template to dropdown with default values { 'react', 'angular', 'iframe' }

TODO: Fix file uploader


1000TurquoisePogs (changes requested):
This needs to be updated against the latest of staging (`git pull --rebase zowe staging`) because it has many conflicts such that I'm not sure how to review this as-is. Many of these files appear to be in common with those already in the codebase, so I suspect a rebase will resolve this down from a 15k line commit to something smaller to review.

PR #289, opened 2020-10-05
Hide app tray + adjust maximize mode when single app mode activates

DivergentEuropeans:
**How to test?**

Set 
`(window as any)['GIZA_SIMPLE_CONTAINER_REQUESTED'] = true;` 
anywhere upon virtual desktop startup, for example in
zlux-app-manager\virtual-desktop\src\app\window-manager\mvd-window-manager\launchbar\launchbar\launchbar.component.ts

To test this PR with an opened app (because the app bar will be hidden, and thus you can't open new apps) I'd recommend using https://github.com/zowe/sample-react-app/pull/30/files or any other auto-save example in the Sample Apps that uses the autosave feature to re-open itself upon Desktop refresh. I don't remember if the re-opening itself relies on auto-save or the desktop auto-save changes and a simple staging build of the Sample React App, but anyway that PR will have both

### DIFF
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
@@ -145,7 +145,8 @@ export class LaunchbarMenuComponent implements MVDHosting.LoginActionInterface{
   ngOnInit(): void {
     this.appKeyboard.keyUpEvent
       .subscribe((event:KeyboardEvent) => {
-        if (event.which === KeyCode.KEY_M) {
+        // TODO: Disable bottom app bar once mvd-window-manager single app mode is functional. Variable subject to change.
+        if (event.which === KeyCode.KEY_M && !(window as any)['GIZA_SIMPLE_CONTAINER_REQUESTED']) {
           this.activeToggle();
         }
     });
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
@@ -10,9 +10,12 @@
   Copyright Contributors to the Zowe Project.
 -->
 
-<div class="launchbar-container launchbar-large" id="container"
-     [style.height]="barSize"
-     (mouseup)="onMouseUpContainer($event)" [class.active]="isActive">
+<div class="launchbar-container launchbar-large" 
+  id="container"
+  [style.height]="barSize"
+  (mouseup)="onMouseUpContainer($event)" 
+  [class.active]="isActive"
+  [style.display]="displayAppBar">
   <rs-com-launchbar-menu [menuItems]="allItems" [theme]="_theme" (itemClicked)="menuItemClicked($event)"  (refreshClicked)="getNewItems()"></rs-com-launchbar-menu>
   
   <div class="applist row" 
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
@@ -15,7 +15,7 @@
   [style.height]="barSize"
   (mouseup)="onMouseUpContainer($event)" 
   [class.active]="isActive"
-  [style.display]="displayAppBar">
+  *ngIf="displayAppBar">
   <rs-com-launchbar-menu [menuItems]="allItems" [theme]="_theme" (itemClicked)="menuItemClicked($event)"  (refreshClicked)="getNewItems()"></rs-com-launchbar-menu>
   
   <div class="applist row" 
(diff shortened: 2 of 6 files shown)


virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
```diff
@@ -10,9 +10,12 @@
   Copyright Contributors to the Zowe Project.
 -->
 
-<div class="launchbar-container launchbar-large" id="container"
-     [style.height]="barSize"
-     (mouseup)="onMouseUpContainer($event)" [class.active]="isActive">
+<div class="launchbar-container launchbar-large" 
+  id="container"
```
lchudinov, line 14: why do we need `id` here?

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
```diff
@@ -10,9 +10,12 @@
...
+  [style.height]="barSize"
+  (mouseup)="onMouseUpContainer($event)" 
+  [class.active]="isActive"
+  [style.display]="displayAppBar">
```
lchudinov, line 18: Could it be possible to it with `*ngIf`?

lchudinov (changes requested):
This works fine. To make the code look better:
* all occurrences of `(window as any)['GIZA_SIMPLE_CONTAINER_REQUESTED'])` must be wrapped into a function/constant
* `*ngIf` must be preferred way to hide DOM elements

DivergentEuropeans:
Copying here from exterior chat for history purposes
Thanks for the review Leonty

> *ngIf must be preferred way to hide DOM elements

You're right, ngIf removes the DOM element completely which could be better for performance, I forgot about this with 'display' attribute

> all occurrences of (window as any)['GIZA_SIMPLE_CONTAINER_REQUESTED']) must be wrapped into a function/constant

I agree with this as well, but I don't think it's useful to do in this PR. That's why I added those TODOs statements, that variable will certainly change. In fact, it may be better to just add a small service for this, but I think the code for that should go when I actually begin working on the standalone mode itself
1:54
What are your thoughts?

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
DivergentEuropeans: Just for formatting because it looked wonky

DivergentEuropeans: Hate having to scroll along the y axis to see something, maybe that's just personal preference

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar-menu/launchbar-menu.component.ts
```diff
@@ -145,7 +145,8 @@ export class LaunchbarMenuComponent implements MVDHosting.LoginActionInterface{
   ngOnInit(): void {
     this.appKeyboard.keyUpEvent
       .subscribe((event:KeyboardEvent) => {
-        if (event.which === KeyCode.KEY_M) {
+        // TODO: Disable bottom app bar once mvd-window-manager single app mode is functional. Variable subject to change.
+        if (event.which === KeyCode.KEY_M && !(window as any)['GIZA_SIMPLE_CONTAINER_REQUESTED']) {
```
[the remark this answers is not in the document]
DivergentEuropeans: See comment

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
lchudinov: Why do we need `id` at all? Do we reference it anywhere?

lchudinov:
It looks good now with `*ngIf`. Let's put `(window as any)['GIZA_SIMPLE_CONTAINER_REQUESTED'])` into a small service in the future PRs.

lchudinov (approved):

virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/launchbar/launchbar.component.html
DivergentEuropeans: Probably QA/automation. It was there so I won't touch it for now

1000TurquoisePogs (changes requested):
Minimize hiding totally works.
But, launch bar hiding doesnt really.
Heres what I saw
1) login to desktop with any custom theme set
2) launchbar exists
3) go into preferences, change the size 
4) still exists
5) open an app
6) go into preferences, change the size
7) NOW the bar is hidden and stays hidden
8) refresh the page
9) launch bar is visible again

The theming is messing with it, probably you are not setting the bar visibility in the right place or in all places.

1000TurquoisePogs:
Disregard by bug report: it was based off of having the toggle for single mode in the wrong place. Placing it early on in the code (as we should) gives the behavior you'd expect.

1000TurquoisePogs (approved):

PR #291, opened 2020-10-09
phase 2 upload

John-A-Davies:
Changes to enable editing of z/OS datasets.


1000TurquoisePogs (commented):
This looks good, but also looks subject to change on what ends up happening with ZSS

John-A-Davies:
References to sessionID are no longer required, as this was a tryout solution which is now superseded.

bootstrap/src/uri/dsm-uri.ts
```diff
@@ -34,9 +34,14 @@ export class DsmUri implements ZLUX.UriBroker {
   omvsSegmentUri(): string {
     return `${this.serverRootUri('omvs')}`;
   }
+
+
```
DivergentEuropeans, line 38: line breaks

bootstrap/src/uri/mvd-uri.ts
```diff
@@ -54,9 +54,15 @@ export class MvdUri implements ZLUX.UriBroker {
   omvsSegmentUri(): string {
     return `${this.serverRootUri('omvs')}`;
   }
+
+
```
DivergentEuropeans, line 58: line breaks

virtual-desktop/package.json
```diff
@@ -13,7 +13,9 @@
     "i18n": "ng-xi18n -p tsconfig.ngx-i18n.json --i18nFormat=xlf  --outFile=messages.xlf && xliffmerge -p xliffmerge.json",
     "lint": "tslint -c tslint.json \"src/**/*.ts\""
   },
-  "dependencies": {},
+  "dependencies": {
+    "@types/qs": "^6.9.3"
+  },
```
DivergentEuropeans, line 18: Do we need these other 2 npm packages?

DivergentEuropeans (commented):
1 question, otherwise lgtm

virtual-desktop/package.json
John-A-Davies: Removed.

PR #305, opened 2021-01-13
add /iframe path for remote iframe plugins

NakulManchanda:
Signed-off-by: Nakul Manchanda <nakul.manchanda@ibm.com>

Resolving gap for:
It will resolve to /iframe path, instead of /web
Standalone/Open in new browser tab option
image

PR #306, opened 2021-01-13
add /iframe path for remote iframe plugins

NakulManchanda:


1000TurquoisePogs (approved):
Looks same as #305 so merging.

PR #361, opened 2021-08-24
Bugfix/display-login-errors

bobadesakshi21:
## Proposed changes

Previously desktop's only error message on failure was saying failure occurred for an array of auth plugins. In order to improve this now the desktop's login UI displays the appropriate error message that is send by the server.

This PR addresses Issue: [[https://jira.rocketsoftware.com/browse/MVD-4506](https://jira.rocketsoftware.com/browse/MVD-4506)]

## Type of change

## PR Checklist
- [x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/virtual-desktop/src/app/authentication-manager/login/login.component.ts
+++ b/virtual-desktop/src/app/authentication-manager/login/login.component.ts
@@ -42,6 +42,9 @@ export class LoginComponent implements OnInit {
   newPassword: string;
   confirmNewPassword: string;
   errorMessage: string;
+  showError: boolean;
+  showErrorDetails: boolean;
+  errorDetails: string;
   loginMessage: string;
   expiredPassword: boolean;
   private passwordServices: Set<string>;
@@ -72,6 +75,9 @@ export class LoginComponent implements OnInit {
     this.enableExpirationPrompt = true;
     this.renewSession = this.renewSession.bind(this);
     this.isIdle = this.isIdle.bind(this);
+    this.showError = false;
+    this.showErrorDetails = false;
+    this.errorDetails = '';
     this.authenticationService.loginScreenVisibilityChanged.subscribe((eventReason: MVDHosting.LoginScreenChangeReason) => {
       switch (eventReason) {
       case MVDHosting.LoginScreenChangeReason.UserLogout:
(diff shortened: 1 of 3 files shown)


bobadesakshi21:
Kindly refer below screenshots
Login_errors
login_errors2

DivergentEuropeans (changes requested):
1. login to Desktop normally
2. log out
3. turn off ZSS
4. try to log in again
wait for error to show, but loading shouldn't stop
5. turn on ZSS again
in 'staging', when you turn it back on, you should be able to log back in successfully but here, because loading doesn't stop, you can't log back in

virtual-desktop/src/app/authentication-manager/login/login.component.ts
```diff
@@ -168,14 +172,33 @@ export class LoginComponent implements OnInit {
...
               }
-              this.errorMessage = this.translation.translate('AuthenticationFailed',
+              this.errorDetails = '';
+              let err;
+              for (let i = 0; i < failedTypes.length; i++) {
+                let plugins = Object.keys(jsonMessage.categories[failedTypes[i]].plugins);
+                for (let j = 0; j < plugins.length; j++) {
+                  err = jsonMessage.categories[failedTypes[i]].plugins[plugins[j]].error;
+                  if(err) {
+                    this.errorMessage = err.message;
```
1000TurquoisePogs, line 189: err.message might not exist. There is no enforcement of what is inside of `error` object.
You should set this.errorMessage to the previous default (`this.translation.translate('AuthenticationFailed', { numTypes: failedTypes.length, types: JSON.stringify(failedTypes) });`) until you find the first object which actually does have an `err.message` to replace the default.

virtual-desktop/src/app/authentication-manager/login/login.component.ts
```diff
@@ -168,14 +172,33 @@ export class LoginComponent implements OnInit {
...
+                    if(!failedPlugins.has(plugins[j])) {
+                      // Appending the error messages and corresponding unique plugin-ids that have errors
+                      this.errorDetails += `${plugins[j]}: ${this.errorMessage}.\n${err.body}\n`;
```
1000TurquoisePogs, line 192: Therefore these may also be undefined. In that case, is there something better you want to put other than
```
org.zowe.zlux.auth.example: undefined
undefined
```
Perhaps "No error message given" and "No error body given"

virtual-desktop/src/app/authentication-manager/login/login.component.ts
```diff
@@ -318,8 +344,27 @@ export class LoginComponent implements OnInit {
               this.expiredPassword = true;
               this.loginMessage = this.translation.translate(PASSWORD_EXPIRED);
             } else {
-              this.errorMessage = this.translation.translate('AuthenticationFailed',
-              { numTypes: failedTypes.length, types: JSON.stringify(failedTypes) });
+              this.errorDetails = '';
+              let err;
+              // Get the server error messages from auth plugins
+              for (let i = 0; i < failedTypes.length; i++) {
```
1000TurquoisePogs, line 350: This looks like code duplication from above. If possible, move the logic into a utility function to prevent bugs in the case that we enhance 1 but not the other.

1000TurquoisePogs (changes requested):
This looks like a great enhancement but I spotted some minor issues. I'm currently making a test build to confirm it works.

1000TurquoisePogs:
Test build works well! No more issues found.

virtual-desktop/src/app/authentication-manager/login/login.component.ts
```diff
@@ -324,50 +297,12 @@ export class LoginComponent implements OnInit {
...
-                this.errorMessage = this.translation.translate('AuthenticationFailed',
-                { numTypes: failedTypes.length, types: JSON.stringify(failedTypes) });
-              }
-              this.password = '';
-            }
-          }
+        if (error.status == HTTP_STATUS_PRECONDITION_REQUIRED) {
+          this.expiredPassword = true;
+          this.loginMessage = this.translation.translate(PASSWORD_EXPIRED);
+        } else if(jsonMessage) {
```
1000TurquoisePogs: I think this breaks the expired password reset ability.
Basically, we need to know what plugins can handle password reset before we can let the user do a password reset. The answer is in the jsonMessage, so you can't if-else this. You need to keep the PASSWORD_EXPIRED check logic within the jsonMessage handling. This might mean that you can't de-duplicate the displayErrorDetails function without knowing which block of code is calling it since there are 2: `attemptLogin` and `ngOnInit`

1000TurquoisePogs (changes requested):

1000TurquoisePogs (approved):
Thanks, this looks like it addresses all my concerns!

DivergentEuropeans (approved):
This needs a CHANGELOG but I will add it in a separate PR

PR #407, opened 2022-01-17
Make pinned plugins function

lchudinov:
## Proposed changes

This PR fixes issues with pinned plugins that occurred because of using new HTTPClient. 

## Type of change

## PR Checklist
- [x] New and existing unit tests pass locally with my changes
- [x] video or image is included if visual changes are made

## Testing

Try to pin/unpin a few plugins.


DivergentEuropeans (approved):
@lchudinov This also works with `res.contents` instead of `res.body.contents` but I trust you're following their API recommendations and this works for me so lgtm!

PR #408, opened 2022-01-20
Fix init bug where a warning was showed when not needed

1000TurquoisePogs:

## Proposed changes

This attempts to fix an issue where the desktop will show an error on the login screen on startup before any user input is sent.
This error is just a way of the server telling the desktop "there is no active session, please login" and so there's no need to show it, but a regression has occurred and so it is shown. This fix attempts to hide it without making real errors invisible.

## Type of change

## PR Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] New and existing unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works, or describe a test method below

## Testing

You should not see any error message on normal startup of the desktop login page, but if you were to type in a wrong password you should still see an error message about login failure.


DivergentEuropeans (approved):
Makes sense

PR #507, opened 2023-03-24
Fix Iframe adapter + replace snapshot code with simpler UI to nearly completely eliminate Desktop multi apps slowdown

DivergentEuropeans:
## Proposed changes

This PR fixes the bug where the isSingleAppMode( ) method no longer works as intended 

(can be seen via opening the Sample Iframe App in standalone mode and seeing that despite in the code, notifications are disabled for isSingleAppMode - which they shouldn't be because we can still use them from the MVD window manager enhancement even in standalone mode - but they are visible in the UI. This is because isSingleAppMode( ) atm always returns TRUE https://github.com/zowe/sample-iframe-app/blob/ca132f90482efba62623bbb61c6498785a7217ab/web/html/index.html#L76)

It also creates isSingleAppModeSimple( ) which means an app is in single app mode + using simple window manager

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)

## PR Checklist
- [ ] If the changes in this PR are meant for the next release / mainline, this PR targets the "staging" branch.
- [ ] My code follows the style guidelines of this project (see: [Contributing guideline](https://github.com/zowe/zlux/blob/master/CONTRIBUTING.md))
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] Relevant update to CHANGELOG.md
- [ ] My changes generate no new warnings

## Testing

So for testing, the changes are basically:
GIZA_SIMPLE_CONTAINER_REQUESTED --> now is only true when we are in simple mode, not just standalone mode
GIZA_PLUGIN_TO_BE_LOADED --> now used in place of GIZA_SIMPLE_CONTAINER_REQUESTED across zlux-app-manager to tell if we are in standalone mode which the incorrectly named GIZA_SIMPLE_CONTAINER_REQUESTED did before

iframe-adapter.js has been updated to reflect above ^

### DIFF
--- a/bootstrap/web/iframe-adapter.js
+++ b/bootstrap/web/iframe-adapter.js
@@ -183,27 +183,54 @@ var ZoweZLUX = {
         pluginDef: undefined,
         launchMetadata: undefined,
 
-        //True - Standalone, False - We are in regular desktop mode
+        //True - Single app mode, False - We are in regular desktop mode
         isSingleAppMode() {
             return new Promise(function(resolve, reject)  {
-                if (window.GIZA_SIMPLE_CONTAINER_REQUESTED) { //Ancient edgecase
+                if (window.top.GIZA_PLUGIN_TO_BE_LOADED) { //Ancient edgecase
+                    resolve(true); //Standalone mode
+                } else { // The below hacky edgecase is to account for timing issues
+                    // let intervalId = setInterval(checkForStandaloneMode, 100);
+                    // function checkForStandaloneMode() {
+                    //     if (ZoweZLUX.iframe.pluginDef) { //If we have the plugin definition
+                    //         clearInterval(intervalId);
+                    //         resolve(false);
+                    //     }
+                    // }
+                    // setTimeout(() => { 
+                    //     clearInterval(intervalId);
+                    //     if (ZoweZLUX.iframe.pluginDef === undefined || null) {
+                    //         resolve(true);
+                    //     } else {
+                    //         resolve(false);
+                    //     }
+                    // }, 1000);
+                    resolve(false);
+                }
+            });
+        },
+
+        //True - Standalone + using simple window manager, False - We are in regular desktop or using the MVD window manager for single app mode
+        isSingleAppModeSimple() {
+            return new Promise(function(resolve, reject)  {
+                if (window.top.GIZA_SIMPLE_CONTAINER_REQUESTED) { //Ancient edgecase
                     resolve(true); //Standalone mode
                 } else {
-                let intervalId = setInterval(checkForStandaloneMode, 100);
-                function checkForStandaloneMode() {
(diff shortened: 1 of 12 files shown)


bootstrap/web/iframe-adapter.js
```diff
@@ -183,27 +183,24 @@ var ZoweZLUX = {
         pluginDef: undefined,
         launchMetadata: undefined,
 
-        //True - Standalone, False - We are in regular desktop mode
+        //True - Single app mode, False - We are in regular desktop mode
         isSingleAppMode() {
             return new Promise(function(resolve, reject)  {
-                if (window.GIZA_SIMPLE_CONTAINER_REQUESTED) { //Ancient edgecase
+                if (window.top.GIZA_PLUGIN_TO_BE_LOADED) { //Ancient edgecase
```
timgerstel, line 190: Why is window.top used in some places and window used in others?

DivergentEuropeans: window.top is used here because zlux-app-manager creates these globals in the window parent. window.top is not used everywhere else in iframe-adapter because as my understanding, the events get emmitted as ZLUX events so people should listen to those anyway, not pollute the parent window further with stuff

DivergentEuropeans:
This PR is 2 parts:

1) Reviewed already internally ✔️  - fixes the Iframe adapter and adds a new feature

2) this commit: https://github.com/zowe/zlux-app-manager/pull/507/commits/0c25ddd6a31300ae0ee5637b371050ecbfb75ac3
- disables existing snapshot feature which happens with every app start and every app refocus
- replaces feature with much lighter UI that can be TODO'd and made better later, or the original snapshotting code can be optimized in a future release!

image

becomes -->

image

results:
- This almost single-handedly makes having 5+ apps in the Desktop usable. I have no issues having 10 Editors open. The performance increase visible from a) app start and b) app refocus is so, so noticeable probably 2x-5x better. 
Simple test: Open 1 Editor. Now open another Editor (you should already see performance increase on start time). Click on the window handle to go back to the first Editor and try to drag this window. Window dragging behaviour will be severely disturbed for 1-3 seconds depending on how bad your network speed is, which is reduced down to about <0.3 seconds with my PR. You can also see the results of this change via the Network logger and the Console logger. Switching between apps in staging will yield more console logs and more Network logs, as our poorly optimized snapshot code attempts to do things it shouldn't be allowed to do. With my PR, there are now no new Network requests or Console logs

 - This fixes an issue with poor Iframe/Window-manager service interaction where If we try to load more than one instance of iframe plugin, this causes more than 3x sources being loaded per plugin (the number increase exponentially with each instance).
Also noticed that if loaded more than one instance of plugin and try focus the instance, it also fires multiple sources being loaded.

zoweIframeAppFocus

struga0258 (approved):

PR #513, opened 2023-06-09
Added support for Node V16 

Priyansh61:
## Proposed changes

- While building on Node V16, the build used to fail.
- The failing apps were
  - Admin Notification App
  - App Prop View
  - System Setting Preference
- Reason behind failing is peer dependency problems with the angular apps.
### Why doesn't the problem happen with node version 14 ?
  - Node version 14 uses npm less than v4-6 In the new version of npm (v7), by default, npm install will fail when it encounters conflicting peerDependencies. It was not like that in previous version of npm where it just used to skip if there was a conflict.
  - Due to the above reason build fails in npm v16.

### How to fix this ?

Most of the problem is caused by the confict of angular dependency v12 and v11, they are not inter-compatible.
So i have updated all the dependency to use angular v12.

### Which dependecies were causing issue ?
- @angular-devkit/build-angular
- Codelyzer
- @rocketsoftware/carbon-component
- Tslint
- @angular/compiler

The above dependencies were causing the problem, I have updated them to use angular v12.

This PR addresses Issue: N.A.
This PR depends upon the following PRs:

## Type of change

## PR Checklist


DivergentEuropeans (commented):
We do have custom Rocket Software Carbon Components, were you able to test @Priyansh61 the Admin Notification App to see if it's still working as intended? I can help you set it up if not

Priyansh61:
@DivergentEuropeans Yeah its working as inteded, here's a proof showing the current admin-notification-component


https://github.com/zowe/zlux-app-manager/assets/96513964/9005fc86-f891-462d-a4d1-0adf50cbc7c3

And I had a doubt, is there a way to directly invoke the app since after installing the plugin it does not show up on the startup Menu, So I had to invoke it from sample angular app. Is there a reason behind this?

Priyansh61:
Closing this one, continuation PR is here https://github.com/zowe/zlux-app-manager/pull/531

PR #531, opened 2023-06-24
Added support for Node V16

Priyansh61:
## Proposed changes

- While building on Node V16, the build used to fail.
- The failing apps were
  - Admin Notification App
  - App Prop View
  - System Setting Preference
- Reason behind failing is peer dependency problems with the angular apps.
### Why doesn't the problem happen with node version 14 ?
  - Node version 14 uses npm less than v4-6 In the new version of npm (v7), by default, npm install will fail when it encounters conflicting peerDependencies. It was not like that in previous version of npm where it just used to skip if there was a conflict.
  - Due to the above reason build fails in npm v16.

### How to fix this ?

Most of the problem is caused by the confict of angular dependency v12 and v11, they are not inter-compatible.
So i have updated all the dependency to use angular v12.

### Which dependecies were causing issue ?
- @angular-devkit/build-angular
- Codelyzer
- @rocketsoftware/carbon-component
- Tslint
- @angular/compiler

The above dependencies were causing the problem, I have updated them to use angular v12.

This PR addresses Issue: N.A.
This PR depends upon the following PRs:

## Type of change

## PR Checklist


DivergentEuropeans:
While you're right that Node 16 does not build correctly, I don't believe this fixes all of them. I get an npm install issue with zlux-platform (zlux-platform isn't buildable, but you can build it indirectly via npm install in zlux-app-manager/bootstrap)

DivergentEuropeans (changes requested):
OH, I see the issue. You're using Node 16.0.0 probably or something very early

This is a great start 👍 but we'll want it to work for the recommended verison of Node. Please use Node 16 LTS 16.16.0 or 16.20.0

Priyansh61:
@DivergentEuropeans I cant understand why it is failing on your system , I actually changed my node version to 16.20.0 and still the build is successfull, so should I upload the package-lock version along with package.json to ensure everything remains the same?

Can you help me with the diagnosis here?

DivergentEuropeans:
@Priyansh61 It's possible this happens because I'm developing on Windows. I'll test this sometime, haven't forgotten about this PR, thank you

DivergentEuropeans:
Issue found, it was old NVM version on Windows

Priyansh61:
@DivergentEuropeans Any other changes to address in this PR ?

PR #602, opened 2024-07-08
js-beautify update

Martin-Zeithaml:
Update `js-beautify` because of black duck.


1000TurquoisePogs:
Due to #601 is this now irrelevant?

Martin-Zeithaml:
My fault, I did not noticed.

PR #613, opened 2024-08-19
Load plugins by entrypoint urls if defined

1000TurquoisePogs:
This PR makes use of https://github.com/zowe/zlux-app-server/pull/308 and https://github.com/zowe/zlux-platform/pull/105 to load the correct entrypoint JS file for native plugins.
It is merely an example as this branch does not have the latest v3 code so it will cause code conflict.


virtual-desktop/src/app/plugin-manager/plugin-factory/angular2/angular2-plugin-factory.ts
```diff
@@ -85,7 +85,8 @@ class SimpleAngularComponentFactory extends ComponentFactory {
 export class Angular2PluginFactory extends PluginFactory {
   private readonly logger: ZLUX.ComponentLogger = BaseLogger;
   private static getAngularModuleURL(pluginDefinition: MVDHosting.DesktopPluginDefinition): string {
-    return ZoweZLUX.uriBroker.pluginResourceUri(pluginDefinition.getBasePlugin(), 'main.js');
+    let entryPoint = pluginDefinition.getBasePlugin().getWebEntryPoint() || 'main.js';
```
DivergentEuropeans: Is this intended behaviour? When webContent is not present, you set entryPoint to undefined (https://github.com/zowe/zlux-platform/pull/105/files#diff-46e106c34def1f9496bcb38ec7bd10d61c9631a6c8ee643a0ce96fbe6b46a8e0R187) , but this line 88 defaults to 'main.js' regardless then

virtual-desktop/src/app/plugin-manager/plugin-factory/react/react-plugin-factory.ts
```diff
@@ -28,7 +28,8 @@ export class ReactPluginFactory extends PluginFactory {
   private readonly logger: ZLUX.ComponentLogger = BaseLogger;
   private static getReactModuleURL(pluginDefinition: MVDHosting.DesktopPluginDefinition): string {
     // TODO: clean this up with .d.ts file
-    return ZoweZLUX.uriBroker.pluginResourceUri(pluginDefinition.getBasePlugin(), 'main.js');
+    let entryPoint = pluginDefinition.getBasePlugin().getWebEntryPoint() || 'main.js';
```
DivergentEuropeans: same here

DivergentEuropeans:
Also, this PR should probably be part of https://github.com/zowe/zlux-app-manager/pull/614 (or both need to go int oV3 branch eventually)

virtual-desktop/src/app/plugin-manager/plugin-factory/angular2/angular2-plugin-factory.ts
1000TurquoisePogs: if webContent is not present, they have an invalid angular plugin and schema validation will fail it upon startup.

if somehow that passes, it becomes no worse than the prior state: main.js is attempted, a 404 is given, and the window in the desktop is blank. no other failure state seems much better given time investment in the unlikely subject

DivergentEuropeans (approved):

PR #614, opened 2024-08-19
New Desktop ID

1000TurquoisePogs:
This PR changes some of the bootstrapping code so that we can load the new desktop with distinct ID.
This PR also allows the old desktop to be switched to via query parameter "?use-v2-desktop=1"


bootstrap/src/bootstrap/bootstrap-manager.ts
```diff
@@ -61,8 +56,18 @@ export class BootstrapManager {
...
 
+      const searchParams = new URLSearchParams(window.location.search);
+      const v2Desktop = desktops.filter((desktop)=> { desktop.getIdentifier() == "org.zowe.zlux.ng2desktop" });
+      const v3Desktop = desktops.filter((desktop)=> { desktop.getIdentifier() == "org.zowe.zlux.ivydesktop" });
+      const useV2Desktop = v2Desktop.length == 1 && searchParams.has("use-v2-desktop") && (searchParams.get("use-v2-desktop") == '1'); 
+      
       if (desktops.length == 0) {
         console.error("ZWED5012E - No desktops available to bootstrap.");
+      } else if (useV2Desktop) {
+        console.error("ZWED5324I - The requested Desktop version (V2) is in maintenance mode. To use the newest Desktop version instead, remove the query parameter 'use-v2-desktop'.");
```
1000TurquoisePogs: i should leave a code comment here: i want this to be colored, but its not really an error.
also change to console.warn

bootstrap/src/assets/i18n/log/messages_en.json
```diff
@@ -41,6 +41,7 @@
     "ZWED5041I":"ectxt=",
     "ZWED5042I":"dispatcher.invokeAction on context %s",
     "ZWED5043I":"RESERVED: MVD standalone container requested with pluginId %s",
+    "ZWED5324I":"RESERVED: The requested Desktop version (%s) is in maintenance mode. To use the newest Desktop version instead, remove the query parameter 'use-v2-desktop'.",
```
DivergentEuropeans, line 44: Going off examples here: https://docs.zowe.org/stable/extend/extend-desktop/mvd-apptoappcommunication/#samples

Maybe let's use format:
useV2Desktop = true | false

bootstrap/src/bootstrap/bootstrap-manager.ts
DivergentEuropeans: So like we touched about on Friday's call, there's utility functions for this @1000TurquoisePogs. The window.location.search stuff is not important enough for duplication but here it is if you want better organization:
`getAllApp2AppArgsFromURL( )` in 
zlux-app-manager\virtual-desktop\src\app\start-url-manager\start-url-manager.service.ts

What is relatively important is there's a registration process for stuff like this that should be followed, see:
zlux-app-manager\bootstrap\src\main.ts
`processApp2AppArgs( )`

URL query params get assigned values in window global object and then those get checked for future reference

virtual-desktop/pluginDefinition.json
```diff
@@ -1,5 +1,5 @@
 {
-  "identifier": "org.zowe.zlux.ng2desktop",
+  "identifier": "org.zowe.zlux.ivydesktop",
```
DivergentEuropeans, line 2: There is a lot of places we now need to change this:
zlux-app-manager\virtual-desktop\src\app\shared\logger.ts
zlux-app-manager\virtual-desktop\src\app\start-url-manager\start-url-manager.service.ts
zlux-app-manager\virtual-desktop\src\app\window-manager\mvd-window-manager\personalization-panel\personalization-panel.component.ts possibly? 
probably others

so do a text search & see if it makes sense to change there too

DivergentEuropeans (changes requested):
At the least identifier needs to be changed in some other places. But some parts of code should be moved around

virtual-desktop/pluginDefinition.json
1000TurquoisePogs: logger & start-url-manager = done
others = no, they seem to have different ids. "org.zowe.zlux.ng2desktop.settings" is not "org.zowe.zlux.ng2desktop". Changing it could have consequences like breaking saved settings.
That may be desirable at some point in the future but I don't see the case for it now.

bootstrap/src/assets/i18n/log/messages_en.json
1000TurquoisePogs: ok

1000TurquoisePogs:
@DivergentEuropeans when you suggested using `getAllApp2AppArgsFromURL( )`, it can't be done.
This is bootstrapping code. That is post-bootstrapping code.

I understand that boostrap interprets query params in a certain way and assigns globals. Let's talk about: why add onto that?

DivergentEuropeans (approved):

PR #617, opened 2024-08-27
Apply angular and library upgrades

1000TurquoisePogs:
Upgrading system-apps to use the latest angular logic, as well as to upgrade all libraries as needed.


DivergentEuropeans (approved):
bootstrap, install and build ✔️ 
admin notif app, install and build ✔️ 
gen, removed
system settings app, install and build ✔️ 
web browser app, install and build ✔️ 
desktop loads, plugins load (VT tested) ✔️

PR #659, opened 2025-02-05
config changes for zlux-editor to work

wadkarsatish:
This PR addresses the issues below.
1. Remove unnecessary webpack module rules. As we are using '@ngtools/webpack' to build the angular contents so no need of 'ts-loader' and 'angular2-template-loader'.
2. The same goes with a rule for loading "html" files. The compilation will be done through '@ngtools/webpack' so no need to explicitly load the html templates.
3. `desktop.ts` added few more library references which is needed for plugins(zlux-editor, file manager)

### DIFF
--- a/virtual-desktop/plugin-config/webpack5.base.js
+++ b/virtual-desktop/plugin-config/webpack5.base.js
@@ -28,11 +28,7 @@ const config = {
     }
   },
   module: {
-    rules: [
-      {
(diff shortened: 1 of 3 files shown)


sakshibobade21 (approved):
Tested these changes, and everything works as expected. Thanks for the PR! Approving it.

PR #685, opened 2026-02-05
Fix app2app bug when open in new tab

1000TurquoisePogs:
This is just a PR to make a test build. Do not merge yet.

A user had an issue where they couldnt use the 'open in new tab' feature of the desktop when their URL already contained some query parameters about app2app. The app2app feature was failing in that new tab.

It appears they had an invalid query parameter set where `?` was seen multiple times.
The line changed here may be the cause, I am not sure yet.

### DIFF
--- a/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/shared/context-utils.ts
+++ b/virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/shared/context-utils.ts
@@ -83,7 +83,11 @@ function openStandalone(item: LaunchbarItem): void {
       window.open(`${location.origin}${ZoweZLUX.uriBroker.pluginResourceUri(item.plugin.getBasePlugin(), pluginWebContent.startingPage)}`);
     }
   } else {
-    window.open(`${location.href}?pluginId=${item.plugin.basePlugin.getIdentifier()}&showLogin=true`);
+    if (location.href.indexOf('?') != -1) {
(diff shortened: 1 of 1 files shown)


virtual-desktop/src/app/window-manager/mvd-window-manager/launchbar/shared/context-utils.ts
```diff
@@ -83,7 +83,11 @@ function openStandalone(item: LaunchbarItem): void {
       window.open(`${location.origin}${ZoweZLUX.uriBroker.pluginResourceUri(item.plugin.getBasePlugin(), pluginWebContent.startingPage)}`);
     }
   } else {
-    window.open(`${location.href}?pluginId=${item.plugin.basePlugin.getIdentifier()}&showLogin=true`);
+    if (location.href.indexOf('?') != -1) {
+      window.open(`${location.href}&pluginId=${item.plugin.basePlugin.getIdentifier()}&showLogin=true`);
+    } else {
+      window.open(`${location.href}?pluginId=${item.plugin.basePlugin.getIdentifier()}&showLogin=true`);
+    }
```
Martin-Zeithaml, line 90: For the readability purposes  I would suggest to use:
```javascript
let concatChar = '?';
if (location.href.indexOf('?') != -1) {
  concatChar = '&';
}
window.open(`${location.href}${concatChar}pluginId=${item.plugin.basePlugin.getIdentifier()}&showLogin=true`);
```

If you read original code, you have to focus to spot a difference between line 87/89.

Martin-Zeithaml (approved):

PR #707, opened 2026-04-09
Replace inline build workflow with reusable zlux-build caller

struga0258:
same as https://github.com/zowe/zlux-app-server/pull/363


1000TurquoisePogs (approved):

1000TurquoisePogs (approved):

1000TurquoisePogs:
This is hung for reasons unclear, and so with James' approval I'm force merging.

